PR #114, opened 2020-02-20
Api Document - 3 api documents (with different stylesheets) are generated

kimeunju108:
Could you please tell me which style you prefer? 
And please feel free to change them.


kimeunju108:
Documents' Location: 
alcor/docs/api_documents/api_1.html
alcor/docs/api_documents/api_2.html
alcor/docs/api_documents/api_3.html

docs/apis/api_1.adoc
```diff
@@ -0,0 +1,390 @@
+= [big blue]*Alcor Controller APIs*
```
xieus: Why do we need 3 API documents? What are the difference of doc 1, 2 and 3?

docs/api_documents/style1.css
```diff
@@ -0,0 +1,408 @@
+/* Asciidoctor default stylesheet | MIT License | http://asciidoctor.org */
+/* Remove comment around @import statement below when using as a custom stylesheet */
```
xieus: Could we simply use an adoc file instead of 10+ files? What do we need html, css etc?

docs/apis/api_1.adoc
```diff
@@ -0,0 +1,390 @@
...
+:stylesheet: style1.css
+
+The followings are the RESTful APIs of Alcor controller system.
+
+[width="100%",cols="25%,15%,39%,21%"]
+|===
+|*Group* |*Method* |*Request*|*Response*
```
xieus: The response column: 

What is Json(Vpc state)? The end user doesn't know about what the exact content a VPC state is. 

A better way is to provide a sample response.

kimeunju108: Hello Liguang,

I agree with yout opinion. I am working on it now.

docs/api_documents/style1.css
kimeunju108: Hello Liguang,

Absolutely, we will use only 1 adoc and maybe 0 or 1 css file eventually. Currently there are many files, but they are candidate files and I will remove once we choose one of them.

docs/apis/api_1.adoc
kimeunju108: Hello Liguang,

That little nicer needs css file, otherwise document is looking a little bit too simple.

xieus: Sure thanks. Let us plan to remove it in the "ultimate" automated version.

docs/api_documents/style1.css
xieus: Cool thanks.

PR #115, opened 2020-02-21
alcor controller logging: change logging directory to home directory

kimeunju108:


src/resources/application-dev.properties
```diff
@@ -7,5 +7,5 @@ apache.kafka.address=172.17.0.1:9092
 #Logging configuration
 logging.level.root=info
 logging.level.org.springframework.web=info
-logging.file.path=/var/log
+logging.file.path=~
```
xieus: Let us also update the instructions on how to change default dictionary of logging file.

kimeunju108: I updated the instruction on wiki

PR #120, opened 2020-02-23
Change default size of log file to 10 MB

xieus:
This PR increases the default size of log file from 1 KB to 10 MB.


xieus:
> Good idea, how about we go bigger like 50 or 100mb? :)

The editor doesn't like that :-)

PR #129, opened 2020-03-17
Replace Redis with Apache Ignite for in memory cache and db

chenpiaoping:
This PR implements the abstraction layer of the in-memory database, which i also call it db cache.  underlying component that really provide cache services that include redis and ignite currently. The ICache interface defines the basic interface for cache operations, such as add, delete, and modify. The cache also supports Transaction,Transaction interface provides transaction-related interfaces. If the underlying cache chooses ignite, it also supports security features and uses SLL to establish a secure link with the ignite server.


src/resources/application.properties
```diff
@@ -9,3 +9,7 @@ logging.level.root=info
 logging.level.org.springframework.web=info
 logging.file.path=.
 logging.type=file
+#Ignite configuration
+#ignite.host=localhost
```
xieus: This is a very good start with the abstraction layer on DB access. 

Couple of suggestions: 
- We likely want to uncomment the ignite configuration. 
- Place ignite cluster on other hosts, and support remote IPs/URLs instead of localhost
- Deploy Ignite clusters as a K8s service

src/com/futurewei/alcor/controller/web/DebugVpcController.java
```diff
@@ -0,0 +1,134 @@
...
+import static org.springframework.web.bind.annotation.RequestMethod.DELETE;
+
+@RestController
+public class DebugVpcController {
+    @Autowired(required = false)
+    private VpcRepository vpcRepository;
+
+    @RequestMapping(
+            method = GET,
+            value = {"/debug/project/{projectid}/vpcs/{vpcid}"})
```
xieus: Good idea of using /debug URL

src/com/futurewei/alcor/controller/db/redis/RedisCache.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.futurewei.alcor.controller.db.ICache;
+import org.springframework.data.redis.core.HashOperations;
+import org.springframework.data.redis.core.RedisTemplate;
+
+import java.util.Map;
+
+public class RedisCache<K, V> implements ICache<K, V> {
+    private HashOperations hashOperations;
+    private String key;
```
xieus: Shall we use other name like "cache_name" to be differentiated with the KEY in K-V?

src/com/futurewei/alcor/controller/db/redis/RedisCache.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public RedisCache(RedisTemplate<K, V> redisTemplate, String key) {
+        hashOperations = redisTemplate.opsForHash();
+        this.key = key;
+    }
+
+    @Override
+    public V get(K var1) {
```
xieus: Let us use "Key" instead of "var"

src/com/futurewei/alcor/controller/db/ignite/IgniteConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.logging.Level;
+
+@Configuration
+@ComponentScan("com.futurewei.alcor.controller.cache")
+@EntityScan("com.futurewei.alcor.controller.cache")
+@ConditionalOnProperty(prefix = "ignite", name = "host")
+public class IgniteConfiguration {
+    private static final Logger logger = LoggerFactory.getLogger();
+
+    @Value("${ignite.host}")
```
xieus: Consider to add SSL support.

src/com/futurewei/alcor/controller/db/ignite/IgniteCache.java
```diff
@@ -0,0 +1,74 @@
...
+
+import javax.cache.Cache;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.stream.Collectors;
+
+
+public class IgniteCache<K, V> implements ICache<K, V> {
+    private static final Logger logger = LoggerFactory.getLogger();
+    private static final int QUERY_PAGE_SIZE = 1000;
```
xieus: Do we know how the page size impacts perf? Could we make it configurable?

src/com/futurewei/alcor/controller/db/ICache.java
```diff
@@ -0,0 +1,18 @@
+package com.futurewei.alcor.controller.db;
+
+import java.util.Map;
+
+public interface ICache<K, V> {
```
xieus: Like this class!

src/com/futurewei/alcor/controller/db/Database.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.ignite.client.IgniteClient;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.data.redis.core.RedisTemplate;
+import org.springframework.stereotype.Component;
+
+@ComponentScan
+@Component
+public class Database {
```
xieus: The class name "Database" is a bit generic. Use something more specific like DB

If different microservices uses different template, how are we going to leverage this class? For example, VPC manager might use vpcTemplate, Subnet Manager might use subnetTemplate etc.

src/com/futurewei/alcor/controller/db/ignite/IgniteConfiguration.java
chenpiaoping: Yes, but SSL handshake failed in thin client mode, while SSL handshake is normal in non-thin client mode. Besides SSL, I also thinking about transaction too.

src/com/futurewei/alcor/controller/db/Database.java
chenpiaoping: Maybe it's time to reconsider a new name.

src/resources/application.properties
```diff
@@ -1,11 +1,18 @@
 #Machine configuration
 alcor.machine.config=classpath:.\\config\\machine.json
 #Connecton configuration
-spring.redis.host=localhost
+spring.redis.host=192.168.131.131
```
xieus: Revert this local change.

src/com/futurewei/alcor/controller/db/ignite/IgniteCache.java
xieus: The static var is gone now. Not sure if we need it.

src/com/futurewei/alcor/controller/cache/repo/ICacheRepository.java
```diff
@@ -16,15 +16,17 @@
 
 package com.futurewei.alcor.controller.cache.repo;
```
xieus: Can you rename this directory from cache.repo to db.repo? Thanks.

src/resources/application.properties
xieus: I think the remaining item is to update the one-box deployment script as well as the k8s yaml file, respectively.

https://github.com/futurewei-cloud/alcor/blob/master/scripts/deploy.sh
https://github.com/futurewei-cloud/alcor/tree/master/kubernetes

How did you test the change locally? deploy a Ignite container and verify locally?

xieus: Let us address this item in next PR.

xieus:
Merge the PR to master. Deployment step to be completed in a separate PR.

PR #137, opened 2020-03-27
add one-box deployment script and k8s yaml file for ignite

chenpiaoping:
This PR adds the yaml files related to the deployment of the ignite cluster in the K8s environment, since the nodes of the ignite cluster need to discover each other, so the ignite Pod needs to communicate to the api-server of K8s to get all the nodes of the ignite cluster. Before creating the ignite Pod, we need to create a ServiceAccount for the ignite and bind it to the Role.


kubernetes/db/ignite-deployment.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: ignite-master
+  labels:
+    app: ignite
+spec:
+  selector:
+    matchLabels:
+      app: ignite
+      role: master
+      tier: backend
+  replicas: 1
```
xieus: Do we test multi replicas? I would suggest to test it out and use 3 as the default replica number.

kubernetes/db/ignite-deployment.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: apps/v1
```
xieus: It is always good practice to write a detailed description for the PR.

PR #158, opened 2020-04-10
Update maven CI workflow to support microservices

xieus:
Update Alcor CI workflow to support build and package of new microservice directories


kimeunju108:
LGTM

PR #171, opened 2020-04-22
Onboard Ignite to VPC Manager and Route Manager

chenpiaoping:
This PR includes the following features:
* Replace the db used by vpc_manager with ignite.
* Replace the db used by route_manager with ignite.
* Move the repo directory under the lib directory to the db directory.


services/route_manager/src/main/resources/application.properties
```diff
@@ -1,11 +1,14 @@
...
+#spring.redis.host=10.103.179.92
+#spring.redis.port=6380
 #apache.kafka.address=172.17.0.1:9092
 #Logging configuration
 logging.level.root=info
 logging.level.org.springframework.web=info
 logging.file.path=.
 logging.type=file
+#Ignite configuration
+#ignite.host=192.168.131.131
```
xieus, line 13: Is this change tested in K8s? I think we need to add ymal files in alcor/kubernetes/services folder.

kevin-zhonghao: Yeah we need to add new ignite ymal files. I haven't test ignite in kubernetes before.

chenpiaoping: Yes, I tested it when I deployed ignite in K8s.

chenpiaoping: I have added some yaml files to ignite and tested the ignite deployment in K8s before.

chenpiaoping: Here are some yaml files for ignite deployment in k8s: https://github.com/futurewei-cloud/alcor/tree/master/kubernetes/db/ignite

PR #175, opened 2020-04-24
[Private IP manager] Skips UTs for mvn packing

chenpiaoping:
This PR includes the following features:
* Skip all test cases when packing.
* Add toString() method for entity classes in Private IP Manager.


services/private_ip_manager/src/main/java/com/futurewei/alcor/privateipmanager/entity/IpAddrAlloc.java
```diff
@@ -64,4 +64,14 @@ public String getState() {
     public void setState(String state) {
         this.state = state;
     }
+
+    @Override
+    public String toString() {
+        return "IpAddrAlloc{" +
```
xieus, line 70: This is a bit manual. 
1) Let us convert the object to a Json string if applicable. 

Try to use GSON or JACKSON. A possible implementation:
    public String toString() {
        Gson gson = new Gson();
        return gson.toJson(this);
    }

2) Make the entity class as a child class, and the parent class has the toString(). So that we don't have to do it in every class.

xieus: Some resource: https://blog.codota.com/how-to-convert-a-java-object-into-a-json-string/

xieus: Do we need to write toString for every class in the future or just a subnet of classes? If the former, we would like to get it a bit automatic for long-term development/debug gain. 

Think about this scenario: if a developer wants to add a new field in the class and forgets to add the new field in the toString() method. Then this new field doesn't show up in the slj4 log and make it harder to debug.

chenpiaoping: Sounds reasonable

xieus:
@chenpiaoping do we still work on this PR? If not significant, we could close it without change, or if you choose to quickly make a change and merge it, that is fine too. Your call.

chenpiaoping:
Since it is not appropriate to skip UTs  during compilation, close this PR.

Issue #150, opened 2020-04-07
[Documentation] Auto-generated Spring REST doc


xieus:
**Basic Requirement**
* Use an automatic way to generate REST API document for API gateway and others microservices listed in this directory: https://github.com/futurewei-cloud/alcor/tree/master/services

**Resource** 
https://github.com/asciidoctor/asciidoctor-intellij-plugin/wiki/Spring-REST-Docs-support

xieus:
@skdong, would you like to take the lead on this item?

skdong:
OK, I will do my best

xieus:
Thank you for the contribution. The task has been assigned.

xieus:
@DavidLiu506 This is a bit old issue which we planned to enable automatic REST documents for our APIs. It has been half done by PR #177 but we didn't test it. 

Could you put this item in your backlog and pick it up when current high-pri items have been completed? Thanks.

PR #176, opened 2020-04-28
Add Apidocs

skdong:
[document] Apis - auto genrate Rest API docs (#150 )

* Add api-gateway, mac-manager,subnet-manager RestDocs documents, including the request and response content in unit  testing, according to the unit test sub-direcotry placed under the adocs direcotry under each microservice

* Add mac-manager,privateip-manager,subnet-manager,vpc-manager Swagger Documents, Swagger Documents include html and pdf two formats ,placed in the swagger dircotry of ech microservice


docs/apis/api-gateway/adocs/admin_get/curl-request.adoc
```diff
@@ -0,0 +1,4 @@
+[source,bash]
```
xieus, line 1: This is definitely a huge PR with 126 files change :-) How do we get those files generated? 

Is it possible that we only check in the codes that could trigger docs auto generation on the compile time, for example, when we do "mvn test & mvn install"

PR #177, opened 2020-04-28
Restii

skdong:
[document] Apis - auto genrate Rest API docs (#150)(#176)

Add api-gateway, mac-manager,subnet-manager,privateip-manger Spring REST Docs generated code


services/mac_manager/src/test/java/com/futurewei/alcor/macmanager/controller/MacControllerTest.java
```diff
@@ -108,7 +111,8 @@ public String createMacRange(MacRange macRange) {
     public void test_index() throws Exception {
         this.mockMvc.perform(get("/start.html"))
                 .andDo(print())
-                .andExpect(status().isOk());
+                .andExpect(status().isOk())
+                .andDo(document("start"));
```
xieus, line 115: should we use "mac_range_post"?

services/api_gateway/src/test/java/futurewei/alcor/apigateway/AlcorApiGatewayApplicationTest.java
```diff
@@ -25,18 +26,35 @@
...
+import static org.springframework.restdocs.webtestclient.WebTestClientRestDocumentation.documentationConfiguration;
+
 import static com.github.tomakehurst.wiremock.client.WireMock.*;
 import static org.assertj.core.api.Assertions.*;
 
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
         properties = {"httpbin=http://localhost:${wiremock.server.port}"})
 @AutoConfigureWireMock(port = 0)
+@AutoConfigureRestDocs(outputDir = "target/generated-snippets")
```
xieus, line 42: If most of the 126 files on the other PR could be found in the target/generated-snippets directory, that would be great. This means that we could safely remove those files in the /docs/apis/ directory as we could always find the latest by running "mvn test". right?

services/api_gateway/pom.xml
```diff
@@ -87,6 +87,17 @@
             <version>0.1.0-SNAPSHOT</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
xieus, line 90: We just merged new UTs to VPC and Route Controllers. Could you please add the docs accordingly. Thank you.

services/api_gateway/src/test/java/futurewei/alcor/apigateway/AlcorApiGatewayApplicationTest.java
xieus: Quick question: are we able to find the html files like alcor/docs/apis/mac-manager/swagger/html/index.html in the target/generated-snippets directory?

skdong: * Yes, we can generate those files under the target / generated-snippets directory by executing mvn test

* The code for the swagger part has not been submitted. Currently, Swagger documentation can be viewed by visiting the /swagger-ui.html path after the microservice is started. Before, I used a separate html generated by a third-party tool. I tried to generate html automatically when I tried mvn test.

services/api_gateway/pom.xml
skdong: Ok let me add it

Now there is a problem, most microservices mvn test need to connect to the database, I currently deploy the database can be executed, but it is best to be able to leave the database

services/mac_manager/src/test/java/com/futurewei/alcor/macmanager/controller/MacControllerTest.java
skdong: * This corresponds to the directory name of the generated document. I think the name and the request are the same for easy search. I can change it if necessary, but it cannot conflict with other requests, otherwise it will overwrite the file

services/api_gateway/src/test/java/futurewei/alcor/apigateway/AlcorApiGatewayApplicationTest.java
xieus: Cool! How can we access the /swagger-ui.html if the microservice is not started?

services/api_gateway/pom.xml
xieus: Is it possible to have a flag which could switch off DB access when generating the docs?

services/api_gateway/src/test/java/futurewei/alcor/apigateway/AlcorApiGatewayApplicationTest.java
skdong: I found some ways to generate swagger html files during mvn test, I am now verifying

services/api_gateway/pom.xml
skdong: The document is generated after the unit test is completed. As long as the unit test does not depend on the database during execution, the document can be generated.

xieus: I see. Please work with @chenpiaoping to see if we can support DB-related UTs by Ignite mock.

PR #180, opened 2020-04-29
[Microservice] Port Manager

chenpiaoping:
This PR contains the following features:
* Basic operations such as get, create, update and delete of port.
* Rollback is supported when the operation of port failed.
* Add async executor to the lib/executor directory.
* Add microservice rest clients to the web/restclient directory.
* Move the entities of each microservice to the web/entity directory so that it can be used when communicating between microservices.


services/port_manager/src/main/java/com/futurewei/alcor/portmanager/exception/AllocateIpAddrException.java
```diff
@@ -0,0 +1,23 @@
...
+        See the License for the specific language governing permissions and
+        limitations under the License.
+*/
+package com.futurewei.alcor.portmanager.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code= HttpStatus.INTERNAL_SERVER_ERROR, reason="Allocate ip address error")
+public class AllocateIpAddrException extends Exception {
```
xieus, line 22: Need an indent

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,572 @@
...
+            ipAddressRollback.putAllocatedIpAddress(result);
+            rollbacks.push(ipAddressRollback);
+        }
+    }
+
+    private String getRangeIdForPort(String vpcId) {
+        return "range1";
+    }
+
+    private void allocateIpAddress(PortState portState, Stack<PortStateRollback> rollbacks) throws Exception {
```
xieus, line 91: The idea of using a rollback stack to keep track of potential rollback call is good. 

The issue, however, is that it might not work in current implementation. If an exception is thrown before rollbacks.push(ipAddressRollback), then the stack doesn't have ipAddressRollback therefore won't be able to trigger rollback during exception handling.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,572 @@
...
+        if (portState.getDnsName() == null) {
+            portState.setDnsName("");
+        }
+
+        if (portState.getDnsAssignment() == null) {
+            portState.setDnsAssignment(new ArrayList<>());
+        }
+    }
+
+    private PortStateJson tryCreatePortState(String projectId, PortState portState, Stack<PortStateRollback> rollbacks) throws Exception {
```
xieus, line 273: The calls into various downstream microservices remain serial and sync. 

One recommended way is to async call into multiple microservices in parallel. See the implementation of Subnet Manager for an example. 

https://github.com/futurewei-cloud/alcor/blob/a4db5c2b368e1a500f4defd62de174d375419392/services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java

lib/src/main/java/com/futurewei/alcor/common/entity/DeviceOwner.java
```diff
@@ -0,0 +1,22 @@
+package com.futurewei.alcor.common.entity;
```
xieus, line 1: It is a good idea to put all common entity classes out of an invidiual microservice. 

Instead of storing them in AlcorCommonLib, we recommend to store the entity classes in AlcorWeb which is under /web directory.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/utils/Ipv4AddrUtil.java
```diff
@@ -0,0 +1,66 @@
...
+        Unless required by applicable law or agreed to in writing, software
+        distributed under the License is distributed on an "AS IS" BASIS,
+        WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+        See the License for the specific language governing permissions and
+        limitations under the License.
+*/
+
+package com.futurewei.alcor.portmanager.utils;
+
+public class Ipv4AddrUtil {
```
xieus, line 19: This is a good util class. Move to common lib so that IP manager could use it as well?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Yes, we can consider it in two cases. First, if an exception occurs before receiving the rest response, there is no need to roll back. Second, if an exception occurs after receiving the rest response, it needs to be rolled back, so we need to ensure that there is no exception between the rest response and rollbacks.push (). In fact, rollbacks.push () is executed after receiving the rest response, so we can assume that there will be no exception between them.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/utils/Ipv4AddrUtil.java
chenpiaoping: Good idea.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Okay, I will do that.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    private void addPortToHost(String hostId) {
+        HostState hostState = getHostState(hostId);
+
+        //FIXME: Add port to Host
+    }
+
+    private void tryCreatePortState(PortState portState, Stack<PortStateRollback> rollbacks) throws Exception {
+        //Verify VPC ID
```
xieus, line 105: We should verify Subnet ID, instead of Vpc Id here.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,421 @@
...
+        if (portState.getMacAddress() == null) {
+            macFuture = AsyncExecutor.execute(macAddressRestWrap::allocateMacAddress, portState);
+        } else {
+            macFuture = AsyncExecutor.execute(macAddressRestWrap::verifyMacAddress, portState);
+        }
+
+        //Verify security group
+
+        //If port binds host, to send it's information to host
+
```
xieus, line 135: Need two more calls: 
(1) One into RouteManager to get the routing rules of the subnet that this port belong to. The list of routing rules is the superset of Neutron router and could include SNAT/DNAT rules.

For now, I would recommend you to look into Subnet implementation regarding how to call RouteManager. 
https://github.com/futurewei-cloud/alcor/blob/2ae412fdbe28296bf92f6001f95989d78730e6ad/services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/service/implement/SubnetServiceImp.java#L120

(2) The other into NodeManager which is under development. 
PR: https://github.com/futurewei-cloud/alcor/pull/185/files

The API you could be interested in is GET /nodes/{nodeid}.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: But port may not have a Subnet ID. In this case, we need to find a suitable subnet for port, and assign an ip address to port from that subnet.

chenpiaoping: @xieus  But port may not have a Subnet ID. In this case, we need to find a suitable subnet for port, and assign an ip address to port from that subnet.

xieus: Correct. We have multiple cases to cover, most of the cases would require to verify subnet ID. 

https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-detail#ports

(1) If a user specifies a subnet ID in the optional fixed_ips field, then we have to verify that subnet Id
(2) If the user doesn't give any fixed_ips, we handle it as you suggested.

chenpiaoping: @xieus  For the first case, I think we need to get the range ID by  Subnet ID (At this point, the Subnet ID will be verified), and then assign an ip address from the ip range. At present, Subnet manager have not provided an interface to obtain the Range ID by the Subnet ID. I discussed this with kevin, and he said he would think about it.

xieus: I think we could make this verification simple by calling the regular Get /subnets/{subnet_id} and the response includes the range ids. @kevin-zhonghao

chenpiaoping: That would be nice，But for the second case, how do i get a suitable Range ID?

xieus: - Simple solution: random pick one subnet 
  (subnet may run out of ips so need retry a few times if such a case is hit)
- Comprehensive solution: Pick subnet based on ip utilization.
We have the ip utilization data for ip range, we could possibly populate that info to VPC manager and pick a subnet with more available ips.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Should I get route info from RouteManager or from SubnetManager? I found that RouteManager only has the interface to get route info by Vpc ID(/ vpcs/ {vpcId} / routes/ {routeId}), and there is no interface to get route info by Subnet ID. However, SubnetManager will store route info, which will be stored in RouteManager,  VpcManager. Will it take up too much memory?

chenpiaoping: For the interaction with node manager, Can i come back to do it after the development of node manager finish?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: I want to deal with the rollbacks of various operations in a unified way by a common method. This method does not need to judge whether the operation needs to be rolled back, and does not need to care about what to be rolled back, this should be concerned by the operation itself. There may be a better way, do you have any good suggestions?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: @xieus To reduce interaction with other microservices, we can add an interface to ip manager. The parameter of this interface is that Vpc ID. This interface will selects an ip range based on Vpc ID and assigns an ip address from that ip range. Is that okay?

xieus: @chenpiaoping Does IP manager has the VPC id?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: @chenpiaoping Regarding NodeManager, PR #185 is under review and should be merged pretty soon so I would recommend to write the codes based on the current APIs. You could comment out for now and test only when NodeManager is merged. Should be soon.

xieus: Regarding routing rules, I suggest to still go to RouteManager as that is the ultimate source of truth for routing information.
I think your point is very good. We possibly need to add a new API in RouteManager to allow query by SubnetId.  Add @kevin-zhonghao for awareness.

xieus: Regarding memory usage of storing routes in SubnetManager and VpcManager that might occupy too much memory,  we haven't seen that case yet so storing routes in the upstream microservices could cut the latency for GET. 

If you find such a case, let me know and we could quickly revisit this design.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: No, not yet.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: okay, xieus.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Okay the db key is range id in IP Manager, and subnet ID is part of the value. If we need to VPC Id there, we expect to see inefficient query if we try to get range by vpc id. Any other thoughts?

xieus: @chenpiaoping Regarding the API to query RangeId by SubnetId, the current Subnet Manager supports that. 

You could use the API 

__GET /project/{projectId}/subnets/{subnetId}__

__Sample Response:__
{
  "subnet": {
    "availability_zone": "string",
    "cidr": "string",
    "description": "string",
    "dhcp_enable": true,
    "dns_list": [
      "string"
    ],
    "gateway_ip": "string",
    "id": "string",
    "ipv4_range_id": "string",
    "ipv6_range_id": "string",
    "mac_address": "string",
    "name": "string",
    "primary_dns": "string",
    "project_id": "string",
    "routes": [
      {
        "associatedTableId": "string",
        "associatedType": "MAIN",
        "description": "string",
        "destination": "string",
        "id": "string",
        "name": "string",
        "priority": 0,
        "project_id": "string",
        "target": "string"
      }
    ],
    "secondary_dns": "string",
    "vpc_id": "string"
  }
}

where ipv4_range_id and ipv6_range_id are what you are looking for :-)

chenpiaoping: Thanks, I found this API. I've submitted a new commit for this.

chenpiaoping: In order to avoid the problem of inefficient query, another idea is that we can create a table to store the mapping between vpc id and rang id, but we need to write this table at the same time when we create or delete an ip range.

xieus:
Question: the latest MockServer will auto generate a new Ingite sub-directory under private_ip_manager after mvn test. What is it used for? We likely need to add the whole sub-directory to .gitignore.

xieus:
After turning on ip manager UTs (commit #5c598ab), some of UTs failed as it tried to connect to Redis.

https://jenkins.alkaidcloud.io/job/alcor-controller-pr-test/298/console

>>> +---------------------------------------------------------------------------------+
>>> Ignite ver. 2.8.0#20200226-sha1:341b01dfd8abf2d9b01d468ad1bb26dfe84ac4f6 stopped OK
>>> +---------------------------------------------------------------------------------+
>>> Grid uptime: 00:00:00.529


[[1;31mERROR[m] [1;31mTests [0;1mrun: [0;1m14[m, Failures: 0, [1;31mErrors: [0;1;31m13[m, Skipped: 0, Time elapsed: 3.438 s[1;31m <<< FAILURE![m - in com.futurewei.alcor.privateipmanager.controller.[1mIpAddrControllerTest[m
[[1;31mERROR[m] Test01_createIpAddrRangeTest  Time elapsed: 0.319 s  <<< ERROR!
org.springframework.web.util.NestedServletException: Request processing failed; nested exception is com.futurewei.alcor.common.db.CacheException: Unable to connect to Redis; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to localhost:6379
	at com.futurewei.alcor.privateipmanager.controller.IpAddrControllerTest.Test01_createIpAddrRangeTest(IpAddrControllerTest.java:63)
Caused by: com.futurewei.alcor.common.db.CacheException: Unable to connect to Redis; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to localhost:6379
	at com.futurewei.alcor.privateipmanager.controller.IpAddrControllerTest.Test01_createIpAddrRangeTest(IpAddrControllerTest.java:63)

chenpiaoping:
services/private_ip_manager/pom.xml


> Question: the latest MockServer will auto generate a new Ingite sub-directory under private_ip_manager after mvn test. What is it used for? We likely need to add the whole sub-directory to .gitignore.



> After turning on ip manager UTs (commit #5c598ab), some of UTs failed as it tried to connect to Redis.
> 
> https://jenkins.alkaidcloud.io/job/alcor-controller-pr-test/298/console
> 
> > > > +---------------------------------------------------------------------------------+
> > > > Ignite ver. 2.8.0#20200226-sha1:341b01dfd8abf2d9b01d468ad1bb26dfe84ac4f6 stopped OK
> > > > +---------------------------------------------------------------------------------+
> > > > Grid uptime: 00:00:00.529
> 
> [�[1;31mERROR�[m] �[1;31mTests �[0;1mrun: �[0;1m14�[m, Failures: 0, �[1;31mErrors: �[0;1;31m13�[m, Skipped: 0, Time elapsed: 3.438 s�[1;31m <<< FAILURE!�[m - in com.futurewei.alcor.privateipmanager.controller.�[1mIpAddrControllerTest�[m
> [�[1;31mERROR�[m] Test01_createIpAddrRangeTest Time elapsed: 0.319 s <<< ERROR!
> org.springframework.web.util.NestedServletException: Request processing failed; nested exception is com.futurewei.alcor.common.db.CacheException: Unable to connect to Redis; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to localhost:6379
> at com.futurewei.alcor.privateipmanager.controller.IpAddrControllerTest.Test01_createIpAddrRangeTest(IpAddrControllerTest.java:63)
> Caused by: com.futurewei.alcor.common.db.CacheException: Unable to connect to Redis; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to localhost:6379
> at com.futurewei.alcor.privateipmanager.controller.IpAddrControllerTest.Test01_createIpAddrRangeTest(IpAddrControllerTest.java:63)

I have deleted the maven-surefire-plugin package in this commit: https://github.com/futurewei-cloud/alcor/pull/180/commits/8c4831b4993c07e4c391fadf8579aaba6a2043ac, do we need to add it back again?

chenpiaoping:
> does this set of UTs have to run in order?

Yes, I annotated @ FixMethodOrder (MethodSorters.NAME_ASCENDING), and all the UT's name is prefixed with TESTxx_, which ensures the order they run.

chenpiaoping:
> We likely need to add the whole sub-directory to .gitignore.

Is the working directory that contains information that ignite nodes need, let me add it to .gitignore.

chenpiaoping:
> After turning on ip manager UTs (commit #5c598ab), some of UTs failed as it tried to connect to Redis.
> 
> https://jenkins.alkaidcloud.io/job/alcor-controller-pr-test/298/console
> 
> > > > +---------------------------------------------------------------------------------+
> > > > Ignite ver. 2.8.0#20200226-sha1:341b01dfd8abf2d9b01d468ad1bb26dfe84ac4f6 stopped OK
> > > > +---------------------------------------------------------------------------------+
> > > > Grid uptime: 00:00:00.529
> 
> [�[1;31mERROR�[m] �[1;31mTests �[0;1mrun: �[0;1m14�[m, Failures: 0, �[1;31mErrors: �[0;1;31m13�[m, Skipped: 0, Time elapsed: 3.438 s�[1;31m <<< FAILURE!�[m - in com.futurewei.alcor.privateipmanager.controller.�[1mIpAddrControllerTest�[m
> [�[1;31mERROR�[m] Test01_createIpAddrRangeTest Time elapsed: 0.319 s <<< ERROR!
> org.springframework.web.util.NestedServletException: Request processing failed; nested exception is com.futurewei.alcor.common.db.CacheException: Unable to connect to Redis; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to localhost:6379
> at com.futurewei.alcor.privateipmanager.controller.IpAddrControllerTest.Test01_createIpAddrRangeTest(IpAddrControllerTest.java:63)
> Caused by: com.futurewei.alcor.common.db.CacheException: Unable to connect to Redis; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to localhost:6379
> at com.futurewei.alcor.privateipmanager.controller.IpAddrControllerTest.Test01_createIpAddrRangeTest(IpAddrControllerTest.java:63)

I tried add maven-surefire-plugin back in the pom.xml file, it's ok in my IDEA env.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: One discipline is that we follow the same workflows for both case (1) and case (2).

You could either go to SubnetManager to verify subnetId or vpcId, retrieve the range id, and then go to IP manager,
OR, build a mapping in IP manager from VpcId to rangeID as well as SubnetId to RangeId, and only need go to IP manager.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: @chenpiaoping I miss your previous comment in this thread. Unifying the rollback is a good idea but depending on how we are going to do it.

Regarding the first case, let us say Service A => Service B, and the modify (POST or PUT or DELETE) request has been sent but no yet received a response, an exception is received at this time. It is still necessary that Service A send roolback to Service B.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Thanks.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+
+public class MockIgniteServer {
+    private static Ignite igniteServer;
+
+    @BeforeClass
+    public static void init() {
+        try {
+            igniteServer = Ignition.start();
```
xieus, line 30: Could we change the default Ignite port to some numbers other than 10800 and use that new default number in the application.properties? This could prevent conflicting with a running Ignite container on the 10800 port.

chenpiaoping: That's a good idea. Let me check the ignite doc.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Oh, this is indeed a big problem. Does subnet manager have the same problem?

chenpiaoping: Submitted a commit for this problem: https://github.com/futurewei-cloud/alcor/pull/180/commits/0a4c96a05592534af52680abe72f9d5d6f364d28

chenpiaoping: https://github.com/futurewei-cloud/alcor/pull/180/commits/0a4c96a05592534af52680abe72f9d5d6f364d28，This commit is going to solve this problem.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Yes, for the case of going to SubnetManager to retrieve a range id,  Since SubnetManager do not know whether there is an available ip address in the range, we may need to go to SubnetManager many times to obtain a available range.

chenpiaoping: @xieus Ip manager provides an interface that randomly assigns ip addresses according to vpcId. Is this solution okay? I'm ready to develop.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Thanks let me check it.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: @chenpiaoping Rethinking the issue, we might have to pick the option of " go to SubnetManager to verify subnetId or vpcId, retrieve the range id, and then go to IP manager". 

The reason is that PortManager needs to visit both RouteManager and IPManager to retrieve routes and to allocate ip, respectively. SubnetManger is the source of truth to retrieve the routes under the subnet and to get the ip range id under the same subnet. 

Therefore, the workflow is like this: 
1. Port Manager => SubnetManager
   1.1 If Subnet Id is known, send and verify subnet Id, retrieve routes under the subnet (as SubnetManager has the latest routes) and to get range Id
   1.2 Otherwise, send vpc id to SubnetManager, SubnetManager picks one subnet, and the rest is the same as Step 1.1

2. Port Manager => Ip Manager, send the range id from Step 1, and allocate ip from the given range.

Currently, Step 1.1 needs a new API from SubnetManager. @kevin-zhonghao and I will work to create one in a parallel PR (#181).

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
```diff
@@ -18,25 +18,35 @@
 import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteException;
 import org.apache.ignite.Ignition;
+import org.apache.ignite.configuration.ClientConnectorConfiguration;
```
xieus, line 21: @chenpiaoping I think the MockIgniteServer and MockRedisServer are very good change which we want to merge to Master quick as other services could start using that. 

Is it possible that we split this PR into two by creating a new PR solely on the Mock change, which is supposed to consist of just a few files? Let me know.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: @xieus  I have submitted the code, but it is a little different from what you described. Can you review it?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
chenpiaoping: If there is no problem with this PR so far, I think we can merge it into the master,  I'll come back and add the part of interaction with route manager and dataplane later.

lib/pom.xml
```diff
@@ -33,6 +33,10 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-thymeleaf</artifactId>
         </dependency>
+        <dependency>
```
xieus, line 36: Does lib needs spring boot?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
chenpiaoping: It's a bit troublesome to split MockIgniteServer from this PR, because many of the UTs of ip manager are based on MockIgniteServer.

services/port_manager/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
```
xieus, line 68: Let us turn on port manger UTs.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/config/ThreadPoolExecutorConfig.java
```diff
@@ -0,0 +1,10 @@
+package com.futurewei.alcor.portmanager.config;
+
+public class ThreadPoolExecutorConfig {
+
+    public static int corePoolSize = 10;
```
xieus, line 5: These are critical parameters. Need comments.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/exception/AllocateIpAddrException.java
```diff
@@ -0,0 +1,22 @@
...
+        WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+        See the License for the specific language governing permissions and
+        limitations under the License.
+*/
+package com.futurewei.alcor.portmanager.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code= HttpStatus.INTERNAL_SERVER_ERROR, reason="Allocate ip address error")
```
xieus, line 21: Like it!

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
xieus: @chenpiaoping This PR is kind of huge with over 100 files change.  Some files are still under review.

Let us check in small change first.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
xieus: Thanks for making the change.

services/port_manager/pom.xml
chenpiaoping: The UT of port manager need to interact with many microservices. If it is opened, the test will fail. Do you have any suggestions?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
xieus: @chenpiaoping I miss your last comment about this PR has dependency on the MockService. In this case, it is okay to keep MockService here. Let me try to review the rest and merge soon then.

lib/pom.xml
chenpiaoping: Oh, it doesn't need it.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/executor/AsyncExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.*;
+import java.util.function.Supplier;
+
+public class AsyncExecutor {
```
xieus, line 28: This could move to /lib once proven to be reliable in IP manager.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
chenpiaoping: Thanks, xieus.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/controller/PortController.java
```diff
@@ -0,0 +1,64 @@
...
+public class PortController {
+    @Autowired
+    PortService portService;
+
+    @PostMapping({"/project/{project_id}/ports", "v4/{project_id}/ports"})
+    @ResponseBody
+    @ResponseStatus(HttpStatus.CREATED)
+    public PortStateJson createPortState(@PathVariable("project_id") String projectId,
+                                         @RequestBody PortStateJson portStateJson) throws Exception {
+        return portService.createPortState(projectId, portStateJson);
```
xieus, line 38: We could verify the input parameters in the controller level.

services/port_manager/pom.xml
xieus: @chenpiaoping SubnetManager UTs has similar issues previously and used Mockito to solve this problem. Check this out:

https://github.com/futurewei-cloud/alcor/blob/master/services/subnet_manager/src/test/java/com/futurewei/alcor/subnet/SubnetControllerTests.java

xieus: @chenpiaoping I will complete reviewing the rest of PR by tomorrow. Meanwhile please use the time to turn on UTs and address existing comments.

chenpiaoping: Okay, thanks xieus.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/utils/BeanUtil.java
```diff
@@ -0,0 +1,49 @@
...
+*/
+package com.futurewei.alcor.portmanager.utils;
+
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+
+@Component
+public class BeanUtil implements ApplicationContextAware {
```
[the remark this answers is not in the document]
chenpiaoping: yes, I thought too, but it didn't work.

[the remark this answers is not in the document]
xieus: Is this because of "@Component" that you will need to launch this bean within the scope of AlcorPortManager?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/executor/AsyncExecutor.java
xieus: @chenpiaoping Could we move the AsycExecutor at this time?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
```diff
@@ -0,0 +1,154 @@
...
+        addIpAddrRollback(new AllocateIpAddrRollback(ipAddressRest), result);
+
+        ipAddrRequests.add(result);
+
+        return ipAddrRequests;
+    }
+
+    public List<IpAddrRequest> verifyIpAddresses(Object args) throws Exception {
+        List<IpAddrRequest> ipAddrRequests = new ArrayList<>();
+        List<PortState.FixedIp> fixedIps = (List<PortState.FixedIp>)args;
```
xieus, line 109: In general, direct type casting is not recommended for downcasting.

Could you try cast() and isInstance() methods for safer downcasting? 

A reference: https://www.baeldung.com/java-type-casting

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        rollbacks.push(rollback);
+    }
+
+    public List<IpAddrRequest> allocateIpAddress(Object args) throws Exception {
+        List<IpAddrRequest> ipAddrRequests = new ArrayList<>();
+        PortState portState = (PortState)args;
+
+        IpAddrRequest result = ipAddressRest.allocateIpAddress(IpVersion.IPV4,
```
xieus, line 91: How about Ipv6? Could we use sth like "portState.IpVersion" instead of fixing with IPv4?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,392 @@
...
+        try {
+            //Verify VPC ID
+            VpcRestWrap vpcRestWrap = new VpcRestWrap(rollbacks);
+            executor.runAsync(vpcRestWrap::verifyVpc, portState);
+
+            IpAddressRestWrap ipAddressRestWrap = new IpAddressRestWrap(rollbacks, portState.getProjectId());
+            if (portState.getFixedIps() == null) {
+                executor.runAsync(ipAddressRestWrap::allocateIpAddress, portState);
+            } else {
+                executor.runAsync(ipAddressRestWrap::verifyIpAddresses, portState.getFixedIps());
```
xieus, line 114: verifyIpAddresses => verifyAndAllocateIpAddress

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,392 @@
...
+            }
+
+            //Wait for all async functions to finish
+            executor.joinAll();
+
+            //Persist portState
+            portRepository.addItem(portState);
+        } catch (Exception e) {
+            /**
+            When an exception occurs, we need to roll back all asynchronous operations,
```
xieus, line 144: Hmm, this is really tricky. We will need a good Timeout story to upper bound the rollback.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -0,0 +1,392 @@
...
+
+    private void UpdateExtraDhcpOpts(PortState portState, PortState portStateOld) {
+
+    }
+
+    private void updatePortToHost(PortState portState) {
+
+    }
+
+    public PortStateJson updatePortState(String projectId, String portId, PortStateJson portStateJson) throws Exception {
```
xieus, line 238: Looks like we'll need a bit more time for updatePortState. It is fine and let us do it in next PR.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/utils/BeanUtil.java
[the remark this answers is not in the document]
chenpiaoping: Oh, maybe, Let me have a try.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/PortService.java
```diff
@@ -0,0 +1,35 @@
...
+package com.futurewei.alcor.portmanager.service;
+
+
+import com.futurewei.alcor.web.entity.port.PortStateJson;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+@Service
+public interface PortService {
```
xieus, line 25: We will need the bulk create API.

web/src/main/java/com/futurewei/alcor/web/rest/AbstractRest.java
```diff
@@ -0,0 +1,31 @@
...
+        distributed under the License is distributed on an "AS IS" BASIS,
+        WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+        See the License for the specific language governing permissions and
+        limitations under the License.
+*/
+package com.futurewei.alcor.web.rest;
+
+import org.springframework.web.client.RestTemplate;
+
+abstract class AbstractRest {
```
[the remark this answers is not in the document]
chenpiaoping: The name your suggest sounds more reasonable, should we also rename all the XXXRest to XXXRestClient?

web/src/main/java/com/futurewei/alcor/web/rest/VpcRest.java
```diff
@@ -0,0 +1,36 @@
...
+        limitations under the License.
+*/
+package com.futurewei.alcor.web.rest;
+
+import com.futurewei.alcor.web.entity.vpc.VpcStateJson;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class VpcRest extends AbstractRest {
```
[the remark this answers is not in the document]
chenpiaoping: ok

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/controller/PortControllerTest.java
```diff
@@ -0,0 +1,335 @@
...
+*/
+package com.futurewei.alcor.portmanager.controller;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.futurewei.alcor.portmanager.config.UnitTestConfig;
+import com.futurewei.alcor.portmanager.repo.PortRepository;
+import com.futurewei.alcor.web.entity.port.*;
+import com.futurewei.alcor.web.entity.mac.*;
+import com.futurewei.alcor.web.entity.vpc.*;
```
xieus, line 24: * vpc/port/mac/ip/mac are in their own sub-dictionary in com.futurewei.alcor.web.entity. 
* subnet and route are still one level up. 

Let us create the same sub-dictionary for subnet and route. Later we could ask other microservices to use the new web classes path and retire the existing ones.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
chenpiaoping: There is no ipVersion field in portState, do you think only ipv4 address is assigned to port by default, or both ipv4 and ipv6 addresses are assigned?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: sounds reasonable.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
```diff
@@ -0,0 +1,154 @@
...
+import com.futurewei.alcor.portmanager.exception.IpVersionInvalidException;
+import com.futurewei.alcor.portmanager.rollback.AllocateIpAddrRollback;
+import com.futurewei.alcor.portmanager.rollback.PortStateRollback;
+import com.futurewei.alcor.portmanager.rollback.ReleaseIpAddrRollback;
+import com.futurewei.alcor.web.rest.SubnetRest;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Stack;
+
+public class IpAddressRestWrap {
```
xieus, line 39: Maybe rename it to "IpAddressRestWrapper". It applies to other classes in the same directory and the directory itself.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/MacAddressRestWrap.java
```diff
@@ -0,0 +1,80 @@
...
+        macState.setMacAddress(portState.getMacAddress());
+
+        addMacAddrRollback(new AllocateMacAddrRollback(macAddressRest), macState);
+
+        return result;
+    }
+
+    public MacStateJson verifyMacAddress(Object args) {
+        //FIXME: Not support yet
+        return null;
```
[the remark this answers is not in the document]
chenpiaoping: that would be nice.

web/src/main/java/com/futurewei/alcor/web/rest/AbstractRest.java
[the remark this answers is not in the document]
xieus: @chenpiaoping Sure. The comment applies to other classes in the same directory and the directory itself (rest=>restClient).

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
xieus: I see. Ipv4v6 is something we need to support as well. Maybe put it in your backlog and do it when you get a chance.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Nice work on support Async with a generic AsyncExecutor class. Like it!

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/controller/PortControllerTest.java
chenpiaoping: subnet and route are still one level up.--->This will cause other microservices to fail to compile. or I need to modify the code of other microservices. Can I do this in the next PR?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: If this mechanism is approved by you, I will move it to the lib directory

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/controller/PortControllerTest.java
xieus: I would suggest you to add two new sub-dictionaries, one for subnet and the other for route, copy those web classes to the new dictionaries, and reference only in PortManager.

We would ask other microservice owners to make the change pointing to the new classes and then remove old/existing ones.

chenpiaoping: Got it.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Sure. Go for it :-)

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
chenpiaoping: I always think wrap or wrapper is not a good name. Do you have any good suggestions?

xieus: A few options: 
* IpAddressRestCoordinator
* IpAddressRestWorker
* IpAddressRestTask

@chenpiaoping It is up to you :-)

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/restwrap/IpAddressRestWrap.java
chenpiaoping: Don't worry, an exception will be thrown if there is a type conversion error, which is exactly what I expected when the type conversion failed.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/controller/PortController.java
```diff
@@ -30,6 +30,16 @@
     @Autowired
     PortService portService;
 
+    /**
+     * Create a port, and call the interfaces of each micro-service according to the
```
xieus, line 34: Very nice comments! Thank you.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/proxy/IpManagerProxy.java
```diff
@@ -84,27 +84,47 @@ private void addIpAddrRollback(AbstractIpAddrRollback rollback, IpAddrRequest ip
         rollbacks.push(rollback);
     }
 
-    public List<IpAddrRequest> allocateIpAddress(Object args) throws Exception {
+    /**
+     * Allocate a random ipv4 and ipv6 address from ip manager service
+     * @param args PortState
+     * @return A list of IpAddrRequest
+     * @throws Exception Rest request exception
```
xieus, line 91: @chenpiaoping, my previous comment may not be 100% clear. I meant that IPv4 only and Ipv4v6 should both be supported, with IPv4 only the default option.

chenpiaoping: I see, should i revert it?

web/src/main/java/com/futurewei/alcor/web/restclient/MacManagerRestClient.java
```diff
@@ -40,11 +40,12 @@ public void releaseMacAddress(String macAddress) throws Exception {
         restTemplate.delete(url);
     }
 
-    public MacStateJson allocateMacAddress(String projectId, String vpcId, String portId) throws Exception {
+    public MacStateJson allocateMacAddress(String projectId, String vpcId, String portId, String mac) throws Exception {
```
xieus, line 43: is this a macAddress or macId?

chenpiaoping: macAddress

chenpiaoping: Let me rename it.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/proxy/IpManagerProxy.java
xieus: Those are good change. Maybe by default we only trigger Ipv4 allocation, but add one more flag to trigger additional Ipv6 allocation. 

Neutron may or may not have this field. Let me quickly check.

chenpiaoping: I think if there is ipv6 subnet in the vpc, the ipv6 address will be assigned.

chenpiaoping: My current implementation may not be very reasonable.

xieus: It is totally fine, we do the development and design in an iterative/agile way: When we find it something not right, we change it quickly. 

My current thinking is: This should be a consistent behavior on VPC level. If a VPC allows Ipv6, every of its Subnets will have Ipv6 ranges therefore trigger Ipv4v6 allocation here.

In our Subnet Object, there is a Ipv4RangeId and a Ipv6RangeId. Therefore our logic could be that if the Ipv6RangeId is not null, we allocate Ipv6 here. Does this make sense?

chenpiaoping: Agree with you, but i think i need to make some changes in ip manager.

chenpiaoping: It's time for me to get off work, will do it tomorrow.

xieus: I see. Might not be a small change considering here requires a full set of UTs. 

Let us comment out Ipv6 allocation part in this PR, and continue to improve it in next PR.

web/src/main/java/com/futurewei/alcor/web/restclient/MacManagerRestClient.java
xieus: No worries. I made changes to expedite the merge.

xieus:
@chenpiaoping UTs are partially turned on with six commented out for now. Please take a look.

PR #193, opened 2020-05-14
[Schema Update] Schema Support for OVS and DHCP

er1cthe0ne:
This is the initial schema update to support OVS. The high-level ideas are:

* Added fields needed to support OVS port configuration update
* Added DHCP configuration
* Keep mizar support for so that it is not a major breaking change for Alcor Control Agent, because ACA still using it for its unit/functional tests

NOTE: It is a draft with a lot of "FIXME" comments for internal discussions.

### DIFF
--- a/schema/proto3/common.proto
+++ b/schema/proto3/common.proto
@@ -36,8 +36,8 @@ enum OperationType {
     DELETE = 3;
     INFO = 4;
     FINALIZE = 5; // FIXME: we will try to the whole port operation in the FINALIZE call
-    CREATE_UPDATE_SWITCH = 6; // FIXME: I don't think we need this for OVS?
-    CREATE_UPDATE_ROUTER = 7;
+    CREATE_UPDATE_SWITCH = 6; // For Mizar only
+    CREATE_UPDATE_ROUTER = 7; // For Mizar only
     CREATE_UPDATE_GATEWAY = 8;
     // FIXME: we will likely need an operation to create/update neighbor info
 }
@@ -66,4 +66,5 @@ enum NetworkType {
     VLAN = 1;
     GRE = 2;
     GENEVE = 3;
+    VXLAN_GPE = 4;
 }
--- a/schema/proto3/common.proto
+++ b/schema/proto3/common.proto
@@ -35,11 +35,14 @@ enum OperationType {
     GET = 2;
     DELETE = 3;
     INFO = 4;
-    NEIGHBOR_INFO = 5; // FIXME: we may need CRUD for NEIGHBOR_INFO
-    CREATE_UPDATE_GATEWAY = 6;
-    FINALIZE = 7;             // For Mizar only
-    CREATE_UPDATE_SWITCH = 8; // For Mizar only
-    CREATE_UPDATE_ROUTER = 9; // For Mizar only
+    NEIGHBOR_CREATE = 5;
+    NEIGHBOR_UPDATE = 6;
+    NEIGHBOR_GET = 7;
+    NEIGHBOR_DELETE = 8;
+    CREATE_UPDATE_GATEWAY = 9;
+    FINALIZE = 10;             // For Mizar only
+    CREATE_UPDATE_SWITCH = 11; // For Mizar only
+    CREATE_UPDATE_ROUTER = 12; // For Mizar only
 }
 
 enum OperationStatus {
--- a/schema/proto3/common.proto
+++ b/schema/proto3/common.proto
@@ -35,11 +35,11 @@ enum OperationType {
     GET = 2;
     DELETE = 3;
     INFO = 4;
-    FINALIZE = 5; // FIXME: we will try to the whole port operation in the FINALIZE call
-    CREATE_UPDATE_SWITCH = 6; // For Mizar only
-    CREATE_UPDATE_ROUTER = 7; // For Mizar only
-    CREATE_UPDATE_GATEWAY = 8;
-    // FIXME: we will likely need an operation to create/update neighbor info
+    NEIGHBOR_INFO = 5; // FIXME: we may need CRUD for NEIGHBOR_INFO
+    CREATE_UPDATE_GATEWAY = 6;
+    FINALIZE = 7;             // For Mizar only
+    CREATE_UPDATE_SWITCH = 8; // For Mizar only
+    CREATE_UPDATE_ROUTER = 9; // For Mizar only
 }
 
 enum OperationStatus {
--- a/schema/proto3/common.proto
+++ b/schema/proto3/common.proto
@@ -26,6 +26,7 @@ enum ResourceType {
     SUBNET = 1;
     PORT = 2;
     SECURITYGROUP = 3;
+    DHCP = 4;
 }
 
 enum OperationType {
@@ -34,10 +35,11 @@ enum OperationType {
     GET = 2;
     DELETE = 3;
     INFO = 4;
-    FINALIZE = 5;
-    CREATE_UPDATE_SWITCH = 6;
+    FINALIZE = 5; // FIXME: we will try to the whole port operation in the FINALIZE call
+    CREATE_UPDATE_SWITCH = 6; // FIXME: I don't think we need this for OVS?
     CREATE_UPDATE_ROUTER = 7;
     CREATE_UPDATE_GATEWAY = 8;
+    // FIXME: we will likely need an operation to create/update neighbor info
 }
 
 enum OperationStatus {
@@ -57,4 +59,11 @@ enum Protocol {
     ICMP = 2;
     HTTP = 3;
     ARP = 4;
+}
+
+enum NetworkType { 
+    VXLAN = 0; // the default type
+    VLAN = 1;
+    GRE = 2;
+    GENEVE = 3;
 }
--- a/schema/proto3/dhcp.proto
+++ b/schema/proto3/dhcp.proto
@@ -19,17 +19,27 @@ syntax = "proto3";
 package alcor.schema;
 
 option java_package = "com.futurewei.alcor.schema";
+option java_outer_classname = "DHCP";
 
-import "vpc.proto";
-import "subnet.proto";
-import "port.proto";
-import "securitygroup.proto";
-import "dhcp.proto";
+import "common.proto";
 
-message GoalState {
-    repeated VpcState vpc_states = 1;
-    repeated SubnetState subnet_states = 2;
-    repeated PortState port_states = 3;
-    repeated SecurityGroupState security_group_states = 4;
-    repeated DHCPState dhcp_states = 5;
+message DHCPConfiguration {
+    int32 version = 1;
+
+    string mac_address = 2;
+    string ipv4_address = 3;
+    string ipv6_address = 4;
+    string port_host_name = 5; // for local DNS response
+
+    message ExtraDhcpOption {
+        string name = 1;
+        string value = 2;
+    }
+
+    repeated ExtraDhcpOption extra_dhcp_options = 6;
+}
+
+message DHCPState {
+    OperationType operation_type = 1;
+    DHCPConfiguration configuration = 2;
 }
--- a/schema/proto3/dhcp.proto
+++ b/schema/proto3/dhcp.proto
@@ -24,7 +24,7 @@ option java_outer_classname = "DHCP";
 import "common.proto";
 
 message DHCPConfiguration {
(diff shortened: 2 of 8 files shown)


schema/proto3/common.proto
```diff
@@ -34,10 +35,11 @@ enum OperationType {
     GET = 2;
     DELETE = 3;
     INFO = 4;
-    FINALIZE = 5;
-    CREATE_UPDATE_SWITCH = 6;
+    FINALIZE = 5; // FIXME: we will try to the whole port operation in the FINALIZE call
+    CREATE_UPDATE_SWITCH = 6; // FIXME: I don't think we need this for OVS?
     CREATE_UPDATE_ROUTER = 7;
```
xieus, line 40: CREATE_UPDATE_SWITCH and CREATE_UPDATE_ROUTER shouldn't be used for OVS scenarios. CREATE_UPDATE_GATEWAY might be. Let us discuss.

schema/proto3/common.proto
```diff
@@ -58,3 +60,10 @@ enum Protocol {
     HTTP = 3;
     ARP = 4;
 }
+
+enum NetworkType { 
+    VXLAN = 0; // the default type
+    VLAN = 1;
+    GRE = 2;
+    GENEVE = 3;
```
xieus, line 68: Is it worth mentioning VXLAN-GPE here?

schema/proto3/dhcp.proto
```diff
@@ -0,0 +1,45 @@
...
+
+package alcor.schema;
+
+option java_package = "com.futurewei.alcor.schema";
+option java_outer_classname = "DHCP";
+
+import "common.proto";
+
+message DHCPConfiguration {
+    int32 version = 1;
```
xieus, line 27: Version shall be needed for other protobuf message as well.

schema/proto3/goalstate.proto
```diff
@@ -24,10 +24,12 @@ import "vpc.proto";
 import "subnet.proto";
 import "port.proto";
 import "securitygroup.proto";
+import "dhcp.proto";
 
 message GoalState {
```
xieus, line 29: We need to discuss the naming of GoalState.

schema/proto3/port.proto
```diff
@@ -26,6 +26,11 @@ import "common.proto";
 message PortConfiguration {
     int32 version = 1;
 
+    // FIXME: the options for NetworkType are:
+    // 1. put in PortConfiguration here since only useful for Port, ACA programming will be direct and easier
+    // 2. put in VPC configuration since it is a "network level" configuration, 
+    //      that mean every port configuration needs to go with VPC config for ACA to look up
+    NetworkType network_type = 13;  // TODO: the ID will be update after discussion
```
xieus, line 33: New field to be placed at the end after discussion.

schema/proto3/port.proto
```diff
@@ -53,16 +58,14 @@ message PortConfiguration {
...
-        string name = 1;
-        string value = 2;
-    }
-
-    HostInfo host_info = 9;
+    HostInfo host_info = 9; // FIXME: will be used to populate neighbor info only
     repeated FixedIp fixed_ips = 10;
     repeated SecurityGroupId security_group_ids = 11;
     repeated AllowAddressPair allow_address_pairs = 12;
-    repeated ExtraDhcpOption extra_dhcp_options = 13;
```
xieus, line 65: Is dhcp options completely covered by the new dhcp.proto

schema/proto3/subnet.proto
```diff
@@ -61,9 +61,9 @@ message SubnetConfiguration {
         string mac_address = 4;
     }
 
-    repeated DnsEntry dns_list = 13;
-    repeated SecurityGroupId security_group_ids = 14;
-    repeated TransitSwitch transit_switches = 15;
+    repeated DnsEntry dns_list = 13; // FIXME: what is this for?
```
xieus, line 64: It is a list of dns name servers associated with the subnet. Default is an empty list.

DNS servers needed to be included by dhcp.proto, right?

schema/proto3/vpc.proto
```diff
@@ -49,7 +53,7 @@ message VpcConfiguration {
 
     repeated SubnetId subnet_ids = 7;
     repeated Route routes = 8;
-    repeated TransitRouter transit_routers = 9;
+    repeated TransitRouter transit_routers = 9; // FIXME: to be removed
```
xieus, line 56: Like it.

schema/proto3/common.proto
er1cthe0ne: Yes, we haven't close done on the Router/DVR design yet. But we can have some initial discussions.

schema/proto3/common.proto
er1cthe0ne: I thought about that, the reasons I didn't put it there are:
1) Current neutron doesn't have support on VXLAN-GPE
2) It is not clear to me on which problem to solve by VLAN-GPE

schema/proto3/dhcp.proto
er1cthe0ne: We have that for PortConfiguration, SubnetConfiguration, VpcConfiguration and now DHCPConfiguration. Do you mean to have version for *every* message including HostInfo or FixedIp under PortConfiguration? I think that would be a good idea.

schema/proto3/port.proto
er1cthe0ne: discussion point: since this is a major revision of the schema, I am proposing to make the fields pretty with the ordering and not to add new fields at the end. Compatibility with the previous release of Alcor or ACA was not a goal in my mind.

schema/proto3/port.proto
er1cthe0ne: yes, that would include DNS server(s) list also.

schema/proto3/subnet.proto
er1cthe0ne: yes, that's cover as dhcp option based on: https://en.wikipedia.org/wiki/Dynamic_Host_Configuration_Protocol

schema/proto3/common.proto
xieus: @haboy52581 These are the fields we discussed.

schema/proto3/common.proto
xieus: > 1. Current neutron doesn't have support on VXLAN-GPE

OVS should support VxLan-GPE already. The OVS commit for GPE support: https://openii.cn/openvswitch/ovs/commit/439f39cb9bd7214d1f70c317420cb2d33556296f

As Neutron doesn't support VxLan-GPE, actually this is a good opportunity. VxLan-GPE is more like a DP term but we could check what's required on CP to support it.

schema/proto3/dhcp.proto
xieus: Versioning to the port level should be sufficient usually. It depends on which new scenarios could be covered for HostInfo and Fixed.

schema/proto3/port.proto
xieus: +1

schema/proto3/common.proto
er1cthe0ne: I have added VxLAN-GPE option now after seeing the advantages: https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/cether/configuration/xe-16/ce-xe-16-book/vxlan-gpe-tunnel.html

Here is my analysis to support vxlan-gpe:
- VM will not need to know that tunnel type is used
- Agent will need to create the additional vxlan-gpe port during setup/init
- Agent will use the new vxlan-gpe port if NetworkType = VXLAN-GPE send down from Controller
- E2E functional testing and automated test code
- I don't see other changes needed unless we start leverage the vxlan-gpe specific features

We can say Alcor will have preliminary or proposed future support for VxLan-GPE. Thoughts?

schema/proto3/common.proto
```diff
@@ -34,10 +35,11 @@ enum OperationType {
     GET = 2;
     DELETE = 3;
     INFO = 4;
-    FINALIZE = 5;
-    CREATE_UPDATE_SWITCH = 6;
-    CREATE_UPDATE_ROUTER = 7;
-    CREATE_UPDATE_GATEWAY = 8;
+    NEIGHBOR_INFO = 5; // FIXME: we may need CRUD for NEIGHBOR_INFO
```
xieus, line 38: Let us close on all the FIXME staff and plan to merge it :-)

schema/proto3/common.proto
xieus: Sounds like a plan. One more question before we call it good: Can VxLan-GPE VNI overlap with "regular" VxLan VNI? Will this cause conflict on the host?

schema/proto3/port.proto
```diff
@@ -24,15 +24,21 @@ option java_outer_classname = "Port";
...
-    string project_id = 2;
-    string network_id = 3;
-    string id = 4;
-    string name = 5;
-    string network_ns = 6;
-    string mac_address = 7;
-    string veth_name = 8;
+    uint32 version = 1;
+
+    // FIXME: the options for NetworkType are:
```
xieus, line 29: Fix FIXME

schema/proto3/common.proto
er1cthe0ne: My understanding is yes, the VxLan-GPE can use the same setup of VNI since it is backward compatible with VxLan. The official VxLan-GPE draft said VxLan-GPE vtep would detect if the other side is VxLan only then it would use VxLan encap header format. However, on the topic of how to detect vtep can support VxLan-GPE or it is VxLan only, the draft said it is out of scope...

schema/proto3/goalstate.proto
er1cthe0ne: I got the idea now and I think we can close this topic.

schema/proto3/common.proto
er1cthe0ne: Fixed.

schema/proto3/port.proto
er1cthe0ne: Fixed.

schema/proto3/port.proto
```diff
@@ -63,10 +59,6 @@ message PortConfiguration {
     repeated FixedIp fixed_ips = 12;
     repeated AllowAddressPair allow_address_pairs = 13;
     repeated SecurityGroupId security_group_ids = 14;
-
-    // FIXME: note that for neutron, it sends down more stuff than this
```
xieus: Let us scope a MVP for basic port programming. We will revisit this part shortly.

schema/proto3/vpc.proto
```diff
@@ -31,10 +31,6 @@ message VpcConfiguration {
     string name = 4;
     string cidr = 5;
     int64 tunnel_id = 6;
-    // FIXME: MTU info would be part VPC right?
```
xieus: That is right. MTU will be on vpc level. 

Does this mean that they will use the same MTU on a region basis and ignore MTU value from Neutron Server.

er1cthe0ne: The MTU setting depends on the physical NIC MTU setting and all the networking hardware needs to support it (e.g. jumbo frame MTU=9000). Therefore, we cannot simply allow a VPC to set certain MTU. More detail here: https://github.com/openstack/neutron/blob/cc69828ff09f5dbc37488f3f56bb045772e35634/doc/source/admin/config-mtu.rst#L7

xieus: Thanks for the clarification. The document makes sense but somehow contradicts with the other document. OpenStack Neutron v2.0 API allows to puts a MTU field when creating a network.

https://docs.openstack.org/api-ref/network/v2/?expanded=create-network-detail#create-network

{
    "network": {
        "name": "sample_network",
        "admin_state_up": true,
        "dns_domain": "my-domain.org.",
        "qos_policy_id": "6a8454ade84346f59e8d40665f878b2e",
        "mtu": 1400
    }
}

er1cthe0ne: Looking deeper into MTU in neutron code, while it is not being used in ovs neutron agent where it is configuring ports created by nova agent, MTU setting are being used in dvr edge router (for source NAT), dvr floating IP management and neutron L3 agent (I am guessing it is because these component will create (e.g. gateway) ports). We will need to look into MTU topic more when we design our router.

xieus: I see. Let us keep this field for now which allows us to revisit for L3 networking staff shortly.

PR #197, opened 2020-05-15
[Port Manager] Add Bulk operation & Integrate with Route/SGM

chenpiaoping:
This PR contains the following features：
* Add validation for PortEntity on controller.
* Support bulk port creation and update.
* Enable Port Manager UTs.
* Support interaction with Route Manager, Security Group Manager and Data-Plane Manager.
* Fix missing fields in PortEntity.


xieus:
@chenpiaoping , you could check the new failures on https://jenkins.alkaidcloud.io/job/alcor-controller-pr-test/367/console

chenpiaoping:
> @chenpiaoping , you could check the new failures on https://jenkins.alkaidcloud.io/job/alcor-controller-pr-test/367/console

sorry, on it.

services/port_manager/pom.xml
```diff
@@ -41,6 +41,12 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.junit.vintage</groupId>
+                    <artifactId>junit-vintage-engine</artifactId>
```
xieus, line 47: @chenpiaoping What does this exclusions block do? It could skip the portion of UTs which potentially fails if enabled, right? Let us fix the actual problem directly.

chenpiaoping: Yes, If we delete it, Mockito won't work, but the reason is unknown. I suspect it conflicts with Mockito.

xieus: Not sure about this. Other services also use Mockito without this exclusion block, and those UTs work okay, for example, vpc/subnet managers.

kimeunju108:
Hello,

I had a same issue. In my case, this is because of Junit 4 and Junit 5 conflict. We are using Junit 5. It was ok to me when I did just FYI,
Add ==> import org.junit.jupiter.api.Test;
Remove ==> import org.junit.Test;

Best,
Eunju Kim

services/port_manager/pom.xml
chenpiaoping: Well, I'll try to find the real reason.

chenpiaoping:
> Junit 4 and Junit 5 conflict



> Remove ==> import org.junit.Test;

@kimeunju108 Thank you for your comment, I tried as you suggested, but it didn't work for port manager.

chenpiaoping:
@xieus 
So far, There are still some questions about port manager:
1. Whether batch deletion is supported, how to support it, two possible solutions:
1) The portId list to be deleted is added in the url parameter, but the length of the url is limited.
2) The portId list to be deleted is added in http body, which does not conform to the specification of Rest and HTTP RFC.
2. Go to route manager to get the route of subnet, but route manager has not provided the corresponding interface.
3. For interacting with dataplane manager, what information should be sent to dataplane manager?

chenpiaoping:
@xieus So far, There are still some questions about port manager:
1. Whether batch deletion is supported, how to support it, two possible solutions:
    a. The portId list to be deleted is added in the url parameter, but the length of the url is limited.
    b. The portId list to be deleted is added in http body, which does not conform to the specification of Rest and HTTP RFC.
2. Go to route manager to get the route of subnet, but route manager has not provided the corresponding interface.
3. For interacting with dataplane manager, what information should be sent to dataplane manager?

services/port_manager/pom.xml
chenpiaoping: 4.29.6. Testing
While it is possible to use JUnit 4 to test Kotlin code, JUnit 5 is provided by default and is recommended. JUnit 5 enables a test class to be instantiated once and reused for all of the class’s tests. This makes it possible to use @BeforeClass and @AfterClass annotations on non-static methods, which is a good fit for Kotlin.

JUnit 5 is the default and the vintage engine is provided for backward compatibility with JUnit 4. If you don’t use it, exclude org.junit.vintange:junit-vintage-engine. You also need to switch test instance lifecycle to "per-class".

To mock Kotlin classes, MockK is recommended. If you need the Mockk equivalent of the Mockito specific @MockBean and @SpyBean annotations, you can use SpringMockK which provides similar @MockkBean and @SpykBean annotations.


seens

chenpiaoping: It seems that the combination of Junit5 and Mockito is not very friendly, so if we need to use Mockito, we'd better use Junit4, so we need junit-vintage-engine to support Junit4? Do I understand correctly?

chenpiaoping: @xieus @kimeunju108

chenpiaoping: Junit 5 requires Java 8 or later, but I found vpc/subnet manager use  java1.7 in pom.xml:
<maven.compiler.source>1.7</maven.compiler.source>
<maven.compiler.target>1.7</maven.compiler.target>
and use Junit4 in UTs:
import org.junit.Test;
but port manager requires Java 8 or later(to support lambda)

xieus:
> @xieus So far, There are still some questions about port manager:
> 
> 1. Whether batch deletion is supported, how to support it, two possible solutions:
>    a. The portId list to be deleted is added in the url parameter, but the length of the url is limited.
>    b. The portId list to be deleted is added in http body, which does not conform to the specification of Rest and HTTP RFC.

We will need to decide whether to support batched deletion based on client behavior and user demands in the near future, i.e., does compute/nova support batch deletion, or does user requires to delete a batch of ports at the same time? There is no clear-cut answer so far. 

If positive, we will want to limit the number of ports in the deletion calls as the nature of port deletion could be coupled tightly with compute resource release. So the url length limit in option a may not be a big issue. What is the limit right now?

> 2. Go to route manager to get the route of subnet, but route manager has not provided the corresponding interface.

@kevin-zhonghao to follow up on this new API request. This new API is expected be available within a week or so.

> 3. For interacting with dataplane manager, what information should be sent to dataplane manager?

Port state bundled with host id/ip, route info, private ip, and virtual mac. DPM API is under development at PR #182.

services/port_manager/pom.xml
xieus: @chenpiaoping By "MockK", do you mean "Mockito"?  Mixing multiple JUnit may not be a good idea as it could cause a bit of headache for compatibility. 

JUnit5 + Mockito2 is possible as suggested here: https://dzone.com/articles/spring-boot-2-with-junit-5-and-mockito-2-for-unit

Maybe worth trying?

xieus:
@chenpiaoping As we discussed earlier, @kevin-zhonghao is trying to unifying all entities classes in multiple microservices by using AlcorWeb entity. Port manager is one of the first microservices completing that task. 

A heads-up: We might need to "upgrade" some entities class used in port Manager, e.g., VPC, as there are recent change in other microservices. After @kevin-zhonghao completes the upgrade, we might need to revisit Port Manager to support those new fields.

xieus:
> > 1. Go to route manager to get the route of subnet, but route manager has not provided the corresponding interface.
> 
> @kevin-zhonghao to follow up on this new API request. This new API is expected be available within a week or so.
> 

Tracking issue: https://github.com/futurewei-cloud/alcor/issues/203

services/port_manager/pom.xml
xieus: Regarding java version, this is a good catch. We should at least use java 8 in vpc/subnet manager at the first place. @kevin-zhonghao

chenpiaoping: > @chenpiaoping By "MockK", do you mean "Mockito"? Mixing multiple JUnit may not be a good idea as it could cause a bit of headache for compatibility.
> 
> JUnit5 + Mockito2 is possible as suggested here: https://dzone.com/articles/spring-boot-2-with-junit-5-and-mockito-2-for-unit
> 
> Maybe worth trying?

yes, I tried that, but it didn't work.

chenpiaoping: spring-boot-starter-test actually contains Junit5 and Junit4 (by including junit-vintage-engine), here is its xml：
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <parent>
    <artifactId>spring-boot-starters</artifactId>
    <version>2.2.6.RELEASE</version>
  </parent>
  <artifactId>spring-boot-starter-test</artifactId>
  <version>2.2.6.RELEASE</version>
  <name>Spring Boot Test Starter</name>
  <description>Starter for testing Spring Boot applications with libraries including
		JUnit, Hamcrest and Mockito</description>
  <url>https://projects.spring.io/spring-boot/#/spring-boot-parent/spring-boot-starters/spring-boot-starter-test</url>
  <organization>
    <name>Pivotal Software, Inc.</name>
    <url>https://spring.io</url>
  </organization>
  <licenses>
    <license>
      <name>Apache License, Version 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0</url>
    </license>
  </licenses>
  <developers>
    <developer>
      <name>Pivotal</name>
      <email>info@pivotal.io</email>
      <organization>Pivotal Software, Inc.</organization>
      <organizationUrl>https://www.spring.io</organizationUrl>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:git://github.com/spring-projects/spring-boot.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/spring-projects/spring-boot.git</developerConnection>
    <url>https://github.com/spring-projects/spring-boot</url>
  </scm>
  <issueManagement>
    <system>Github</system>
    <url>https://github.com/spring-projects/spring-boot/issues</url>
  </issueManagement>
  <dependencies>
    <dependency>
      <artifactId>spring-boot-starter</artifactId>
      <version>2.2.6.RELEASE</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <artifactId>spring-boot-test</artifactId>
      <version>2.2.6.RELEASE</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <artifactId>spring-boot-test-autoconfigure</artifactId>
      <version>2.2.6.RELEASE</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>com.jayway.jsonpath</groupId>
      <artifactId>json-path</artifactId>
      <version>2.4.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>jakarta.xml.bind</groupId>
      <artifactId>jakarta.xml.bind-api</artifactId>
      <version>2.3.3</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.junit.jupiter</groupId>
      <artifactId>junit-jupiter</artifactId>
      <version>5.5.2</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.junit.vintage</groupId>
      <artifactId>junit-vintage-engine</artifactId>
      <version>5.5.2</version>
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <artifactId>hamcrest-core</artifactId>
          <groupId>org.hamcrest</groupId>
        </exclusion>
      </exclusions>
    </dependency>
    <dependency>
      <groupId>org.mockito</groupId>
      <artifactId>mockito-junit-jupiter</artifactId>
      <version>3.1.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.assertj</groupId>
      <artifactId>assertj-core</artifactId>
      <version>3.13.2</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.hamcrest</groupId>
      <artifactId>hamcrest</artifactId>
      <version>2.1</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.mockito</groupId>
      <artifactId>mockito-core</artifactId>
      <version>3.1.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.skyscreamer</groupId>
      <artifactId>jsonassert</artifactId>
      <version>1.5.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.springframework</groupId>
      <artifactId>spring-core</artifactId>
      <version>5.2.5.RELEASE</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.springframework</groupId>
      <artifactId>spring-test</artifactId>
      <version>5.2.5.RELEASE</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.xmlunit</groupId>
      <artifactId>xmlunit-core</artifactId>
      <version>2.6.4</version>
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <artifactId>jaxb-api</artifactId>
          <groupId>javax.xml.bind</groupId>
        </exclusion>
      </exclusions>
    </dependency>
  </dependencies>
</project>

chenpiaoping: with junit-vintage-engine maybe ok? I found some examples like this on the Internet.

chenpiaoping: Since the scope of the built-in junit-vintage-engine is compile, When I change it to test, Mockito is OK, but we don't want to modify the pom.xml file of springboot, so we include junit-vintage-engine through exclusions. The question is why Mockito doesn't work when scope is compile.

chenpiaoping:
> @chenpiaoping As we discussed earlier, @kevin-zhonghao is trying to unifying all entities classes in multiple microservices by using AlcorWeb entity. Port manager is one of the first microservices completing that task.
> 
> A heads-up: We might need to "upgrade" some entities class used in port Manager, e.g., VPC, as there are recent change in other microservices. After @kevin-zhonghao completes the upgrade, we might need to revisit Port Manager to support those new fields.

Checked. no effect on port manager.

chenpiaoping:
> We will need to decide whether to support batched deletion based on client behavior and user demands in the near future, i.e., does compute/nova support batch deletion, or does user requires to delete a batch of ports at the same time? There is no clear-cut answer so far.
> 
> If positive, we will want to limit the number of ports in the deletion calls as the nature of port deletion could be coupled tightly with compute resource release. So the url length limit in option a may not be a big issue. What is the limit right now?

neutron does not support batch deletion. the limit depends on the implementation of the http client.

services/port_manager/pom.xml
chenpiaoping: The starter also brings the vintage engine so that you can run both JUnit 4 and JUnit 5 tests. If you have migrated your tests to JUnit 5, you should exclude JUnit 4 support, as shown in the following example:<dependency>     <groupId>org.springframework.boot</groupId>     <artifactId>spring-boot-starter-test</artifactId>     <scope>test</scope>     <exclusions>         <exclusion>             <groupId>org.junit.vintage</groupId>             <artifactId>junit-vintage-engine</artifactId>         </exclusion>     </exclusions> </dependency>
--


https://docs.spring.io/spring-boot/docs/2.2.7.RELEASE/reference/html/spring-boot-features.html#boot-features-testing

chenpiaoping: The starter also brings the vintage engine so that you can run both JUnit 4 and JUnit 5 tests. If you have migrated your tests to JUnit 5, you should exclude JUnit 4 support, as shown in the following example:
<dependency>     <groupId>org.springframework.boot</groupId>     <artifactId>spring-boot-starter-test</artifactId>     <scope>test</scope>     <exclusions>         <exclusion>             <groupId>org.junit.vintage</groupId>             <artifactId>junit-vintage-engine</artifactId>         </exclusion>     </exclusions> </dependency>

chenpiaoping: https://docs.spring.io/spring-boot/docs/2.2.7.RELEASE/reference/html/spring-boot-features.html#boot-features-testing

The starter also brings the vintage engine so that you can run both JUnit 4 and JUnit 5 tests. If you have migrated your tests to JUnit 5, you should exclude JUnit 4 support, as shown in the following example:
--

chenpiaoping: <dependency>
    <artifactId>spring-boot-starter-test</artifactId>
    <scope>test</scope>
    <exclusions>
        <exclusion>
            <groupId>org.junit.vintage</groupId>
            <artifactId>junit-vintage-engine</artifactId>
        </exclusion>
    </exclusions>
</dependency>

chenpiaoping: mock service can work, that is the reason why vpc/subnet manager is ok.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -115,57 +162,53 @@ public PortStateJson createPortState(String projectId, PortStateJson portStateJs
...
+        AsyncExecutor executor = new AsyncExecutor();
 
-            //Verify Binding Host ID
-            if (portState.getBindingHostId() != null) {
-                NodeManagerProxy nodeManagerProxy = new NodeManagerProxy(rollbacks);
-                nodeManagerProxy.verifyHost(portState.getBindingHostId());
+        try {
+            for (PortState portState: portStateBulkJson.getPortStates()) {
+                portState.setProjectId(projectId);
+                createPortStateAsync(portState, executor, rollbacks);
```
xieus, line 199: By sharing the same rollbacks stack, will one port creation failure cause rollback of other ports in the same request?

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/config/UnitTestConfig.java
```diff
@@ -82,4 +85,42 @@
             "        \"allowed_address_pairs\":[{\"ip_address\":\"" + ip2 + "\", \"mac_address\":\"" + mac1 + "\"}]\n" +
             "    }\n" +
             "}";
+    public static String createPortBulk = "{\n" +
```
xieus, line 88: Bulk APIs are very fancy. Let us add some UTs to cover them :-)

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Yes, I think for batch creation, if a port creation fails, all operations should be rolled back.

services/port_manager/pom.xml
xieus: @chenpiaoping If you click https://github.com/futurewei-cloud/alcor/pull/197/checks?check_run_id=684996404, you could find the current merge results shows as follows:

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.15 s - in com.futurewei.alcor.portmanager.PortManagerApplicationTests
2020-05-18 11:04:16.840  INFO 3974 --- [extShutdownHook] o.s.s.concurrent.ThreadPoolTaskExecutor  : Shutting down ExecutorService 'applicationTaskExecutor'
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0

The goal for us is to enable on UTs for every microservice. Port Manager is so important that we definite need it on.

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/config/UnitTestConfig.java
chenpiaoping: There are two: createPortBulkTest and updateMacAddressAndFixedIpsBulkTest

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: This is right from customer behavior perspective. On the API reliability side, it may not fragile.

Can we add one optional parameter for the API, e.g. allow_partial with default value false. Then the default value is what we are right now. When it is set to true, we will return something all port states, including created/updated port states and failed port with exception.

services/port_manager/pom.xml
chenpiaoping: Oh, let me check, is there a problem with these outputs?

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/config/UnitTestConfig.java
xieus: Oh I see. I didn't find it in this PR. Are they checked in previously?

xieus: Never mind. Find it now.

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/controller/PortControllerTest.java
```diff
@@ -243,19 +256,48 @@ public void createPortWithoutMacAddressTest() throws Exception {
...
+        Mockito.when(macManagerRestClient.allocateMacAddress(UnitTestConfig.projectId, UnitTestConfig.vpcId, UnitTestConfig.portId2, null))
+                .thenReturn(newMacStateJson(UnitTestConfig.portId2, UnitTestConfig.mac2));
+
+        Mockito.doNothing().when(portRepository);
+
+        this.mockMvc.perform(post(createPortBulkUrl).contentType(MediaType.APPLICATION_JSON)
+                .content(UnitTestConfig.createPortBulk))
+                .andDo(print())
+                .andExpect(status().isCreated())
+                .andExpect(MockMvcResultMatchers.jsonPath("$.ports[0].id").value(UnitTestConfig.portId))
```
xieus, line 294: Remove line.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: okay, I know what you mean.

services/port_manager/pom.xml
xieus: Only one test case is on, which is PortManagerApplicationTests. All the good test cases in PortControllerTest.java are not actually running. That is why I am trying to turn them all on.

chenpiaoping: I can see outputs of all UTs from section Build with Maven.
The snippet:
      HTTP Method = POST
      Request URI = /project/3dda2801-d675-4688-a63f-dcda8d327f50/network_segment_ranges
       Parameters = {}
          Headers = [Content-Type:"application/json;charset=UTF-8"]
             Body = {"network_segment_range":{"project_id":"3dda2801-d675-4688-a63f-dcda8d327f50","id":"9192a4d4-ffff-4ece-b3f0-8d36e3d86000","name":"test_segment","description":"","network_type":"vlan"}}
    Session Attrs = {}

Handler:
             Type = com.futurewei.alcor.vpcmanager.controller.SegmentRangeController

xieus: That is VPC manager UT.

chenpiaoping: sorry, I read it wrong.

lib/src/main/java/com/futurewei/alcor/common/db/repo/ICacheRepository.java
```diff
@@ -28,5 +29,7 @@
 
     void addItem(T newItem) throws CacheException;
 
+    //void addItems(List<T> items) throws CacheException;
```
xieus, line 32: Btw, so Ignite doesn't support a bulk insertion?

services/port_manager/pom.xml
xieus: Just checking in to see if you are still blocked on this.

chenpiaoping: I'm already on security group. back soon.

xieus: It is okay. Let us know once it is ready to review then @chenpiaoping

chenpiaoping: ok

lib/src/main/java/com/futurewei/alcor/common/db/repo/ICacheRepository.java
chenpiaoping: Ignite support a bulk insertion,  but if I add this line, all the classes that implement ICacheRepository fail to compile.

services/port_manager/pom.xml
chenpiaoping: The problem has been solved.

xieus: Cool thanks! What was the catch?

chenpiaoping: org.junit.Test belongs to junit4, after excluding junit4, we should not use it any more, or the UTs would not be executed at compile time.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: What does the responsed data look like when we support allow_partial?

chenpiaoping: Supporting allow_partial under the current rollback framework may be a bit troublesome, because we don't know which operations need to be rolled back when an exception occurs.

chenpiaoping: @xieus Regardless of this question, I think this PR is ready to be reviewed.

xieus: okay then. I will resume the review.

lib/src/main/java/com/futurewei/alcor/common/db/repo/ICacheRepository.java
xieus: In that case, we could in a project-wide search for ICacheRepository. Once we turn on the line, we add one missing method for each occurrence and mark them "Not Implemented".

This is not a urgent matter, but we will need bulk insertion eventually.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -94,6 +96,51 @@ private void addPortToHost(String hostId) {
         //FIXME: Add port to Host
     }
 
+    private void createPortStateAsync(PortState portState, AsyncExecutor executor, Stack<PortStateRollback> rollbacks) throws Exception {
```
xieus, line 99: As the requested RouteManager API is ready (ref:https://github.com/futurewei-cloud/alcor/issues/203), we should be able to add integration with Route Manager now.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -247,6 +290,68 @@ private void updatePortToHost(PortState portState) {
 
     }
 
+    private void updatePortStateAsync(PortState portState, PortState oldPortState, AsyncExecutor executor,
```
xieus, line 293: Is update port implementation completed?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -247,6 +290,68 @@ private void updatePortToHost(PortState portState) {
...
+                                      Stack<PortStateRollback> rollbacks) throws Exception {
+        //Update mac_address
+        String macAddress = portState.getMacAddress();
+        String oldMacAddress = oldPortState.getMacAddress();
+        if (macAddress != null && !oldMacAddress.equals(macAddress)) {
+            MacManagerProxy macManagerProxy = new MacManagerProxy(rollbacks);
+            executor.runAsync(macManagerProxy::releaseMacAddress, oldPortState);
```
xieus, line 300: We should allocate mac address before releasing the old one. Otherwise, it is possible that after releasing the old mac, Port Manager instance crashes which causes loss of old mac but no new mac is allocated. In the worse case, the old mac address is allocated by some other tenant.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -247,6 +290,68 @@ private void updatePortToHost(PortState portState) {
...
+            List<PortState.FixedIp> addFixedIps = fixedIpsCompare(fixedIps, oldFixedIps);
+            List<PortState.FixedIp> delFixedIps = fixedIpsCompare(oldFixedIps, fixedIps);
+
+            if (delFixedIps.size() > 0) {
+                executor.runAsync(ipManagerProxy::releaseIpAddressBulk, delFixedIps);
+            }
+
+            if (addFixedIps.size() > 0) {
+                executor.runAsync(ipManagerProxy::allocateFixedIpAddress, addFixedIps);
+            }
```
xieus, line 333: Same there. Allocate before release.

Also, could you give a simple comment or example to explain addFixedIps and delFixedIps to increase code readability?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -267,84 +372,70 @@ public PortStateJson updatePortState(String projectId, String portId, PortStateJ
...
-                }
-            }
+            //Wait for all async functions to finish
+            executor.joinAll();
 
-            //Update fixed_ips
-            List<PortState.FixedIp> fixedIps = portState.getFixedIps();
-            IpManagerProxy ipManagerProxy = new IpManagerProxy(rollbacks, projectId);
+            //Persist the new configuration of port to the db
+            portRepository.addItem(oldPortState);
```
xieus, line 389: newPortState?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -267,84 +372,70 @@ public PortStateJson updatePortState(String projectId, String portId, PortStateJ
...
-            if (portRepository.findItem(portId) == null) {
+            PortState oldPortState = portRepository.findItem(portId);
+            if (oldPortState == null) {
                 throw new PortStateNotFoundException();
             }
 
-            portState.setProjectId(projectId);
-
-            //Update mac_address
+            updatePortStateAsync(portState, oldPortState, executor, rollbacks);
```
xieus, line 383: The update logic is quite tricky. Do we have sufficient UTs to cover those corner cases?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
kevin-zhonghao: Piaoping ~ Have you completed the implementation of port update? When I tested it before, it seemed that the port update API function was not fully completed.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
```diff
@@ -0,0 +1,600 @@
...
+			"ip_address": "11.11.11.100"
+		}],
+		"allowed_address_pairs": [{
+			"ip_address": "11.11.11.101",
+			"mac_address": "00:01:6C:06:A6:29"
+		}],
+		"extra_dhcp_opts": null,
+		"security_groups": ["3d53801c-32ce-4e97-9572-bb966f4d45ca"],
+		"binding:host_id": null,
+		"binding:profile": null,
```
xieus, line 138: For port binding properties, could we add minimum set of validation, for example, only allow specific given value to pass?

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
```diff
@@ -0,0 +1,600 @@
...
+		"binding:vif_details": null,
+		"binding:vif_type": null,
+		"binding:vnic_type": null,
+		"network_ns": null,
+		"dns_name": null,
+		"dns_assignment": null
+	}
+}
```
xieus, line 146: We don't have created_at and updated_at value. Refer to latest VPC/Subnet manager implementation, which just added those fields.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
```diff
@@ -0,0 +1,600 @@
...
+		"extra_dhcp_opts": null,
+		"security_groups": ["3d53801c-32ce-4e97-9572-bb966f4d45ca"],
+		"binding:host_id": null,
+		"binding:profile": null,
+		"binding:vif_details": null,
+		"binding:vif_type": null,
+		"binding:vnic_type": null,
+		"network_ns": null,
+		"dns_name": null,
+		"dns_assignment": null
```
xieus, line 144: Some missing fields to be added:
* dns_domain
* ip_allocation 
* port_security_enabled
* qos_network_policy_id
* qos_policy_id
* revision_number
* resource_request
* tags
* uplink_status_propagation
* mac_leaning_enabled

References: https://docs.openstack.org/api-ref/network/v2/?expanded=show-port-details-detail#show-port-details

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Since both allocate and release are asynchronous, the order of allocate and release cannot be guaranteed.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: no, newPortState not have complete fields.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Good point. But we still needs to address this issue. Two options:
* Based on the different exceptions, Port Manager handles it differently. For example, if allocate fails and release succeeds, we need to retry allocation heavily; if allocate succeeds and release fails, we need to verify that the mac is still own by existing port then return new mac.
* MacManager provides one API to "replace" mac for a port in one "transaction". This appears to be cleaner and easier for clients such as PortManager.

What do you think of?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Same replacement API might be needed for PIP manager as well.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Sorry I don't quite follow this. Why do we want to persist the "oldPortState" in an update call? Can yo u please elaborate?

chenpiaoping: Because some fields in PortState are generated dynamically during create, in addition, only part of the fields are allowed to be updated during update, so the fields in portState may be incomplete.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: Mac Manager to provide an update interface would be nice, at least reducing the number of interactions with it. But this exceptions will still exist(for example, allocate  mac address succeeded, but allocate ip address failed), if we handle this exception in the main flow of CRUD processing, it may be a bit complicated, and there is no 100% guarantee that the exception will be handled correctly (for example, what should we do if we try again but fails?). The simple handling is that once an exception occurs, the whole operation is considered to have failed, and the rollback mechanism is used to undo all operations before the failure.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: @kevin-zhonghao only ip addresses can be updated currently.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
```diff
@@ -0,0 +1,600 @@
...
+			"ip_address": "11.11.11.101",
+			"mac_address": "00:01:6C:06:A6:29"
+		}]
+	}
+}
+
+Response:
+{
+	"port": {
+		"fastPath": false,
```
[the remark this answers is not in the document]
chenpiaoping: Oh, let me check why it became fastPath.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
chenpiaoping: Yes, let me add some logic of validation checking in the PortController.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
chenpiaoping: ok

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: > Mac Manager to provide an update interface would be nice, at least reducing the number of interactions with it. But this exceptions will still exist(for example, allocate mac address succeeded, but allocate ip address failed), 

Does this comment apply to the replacement API? The replacement API offers atomic operation, either success together or failure together, no mix of the two. 

> if we handle this exception in the main flow of CRUD processing, it may be a bit complicated, and there is no 100% guarantee that the exception will be handled correctly (for example, what should we do if we try again but fails?). The simple handling is that once an exception occurs, the whole operation is considered to have failed, and the rollback mechanism is used to undo all operations before the failure.

Correct. The rollback mechanism is intact to protect resources across microservices as described, but not within the same microservices therefore we need the upstream service to provide "replacement" API so that client don't need to worry about those complicated rollback mechanism.

xieus: If no objection, let us proceed to provide two "replacement" APIs, one in Mac Manager, and one in IP Manager then. Thoughts?

chenpiaoping: Agreed. No problem.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
xieus: @chenpiaoping Heads-up: there is another PR #220 to be merged pretty soon which could cause merge conflict with this one. Let us plan to address the remaining issues and merge this one shortly, or more merge conflicts could be introduced by other PRs.

chenpiaoping: ok

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
chenpiaoping: @RequestMapping(
            method = GET,
            value = {"/subnets/{subnetId}/routes/{routeId}"})
    public RouteWebJson getRuleBySubnetId(@PathVariable String subnetId, @PathVariable String routeId) throws Exception {

        RouteEntity routeEntity = null;

        try {
            RestPreconditionsUtil.verifyParameterNotNullorEmpty(subnetId);
            RestPreconditionsUtil.verifyParameterNotNullorEmpty(routeId);

            routeEntity = this.routeDatabaseService.getByRouteId(routeId);
        } catch (ParameterNullOrEmptyException e) {
            //TODO: REST error code
            throw new Exception(e);
        }

        if (routeEntity == null) {
            //TODO: REST error code
            return new RouteWebJson();
        }

        return new RouteWebJson(routeEntity);

    }
Does port manager get the route through this interface? But it requires subnetId and routeId. I thought it just subnetId. Is that right?

xieus: @chenpiaoping 

Here is the link to API: https://github.com/futurewei-cloud/alcor/blob/26033d0c5aeacd9502e3b0c6bd8c518f8eb7da40/services/route_manager/src/main/java/com/futurewei/alcor/route/controller/RouteController.java#L127

chenpiaoping: > @chenpiaoping
> 
> Here is the link to API:
> 
> https://github.com/futurewei-cloud/alcor/blob/26033d0c5aeacd9502e3b0c6bd8c518f8eb7da40/services/route_manager/src/main/java/com/futurewei/alcor/route/controller/RouteController.java#L127

Yes, this is the same API as the one I posted above. I mean, this API needs to provide two parameters, subnetId and routeId, but port manager can't get the routeId, dose port manager need to provide routeId?

xieus: You are right, Piaoping. Reopen issue #203 as unresolved, and add @kevin-zhonghao to take a further look. 

Let us do this quickly. @chenpiaoping, can you add the API in RouteController.java without the implementation, just to unblock yourself to write necessary codes in Port Manager. @kevin-zhonghao or I will implement the API early next week.

chenpiaoping: Don't worry, this won't block me. Just raised it up here.

xieus:
@chenpiaoping I am on this PR. Could you update the description/subject of the PR as it slowly becomes a humongous PR with quite a few new features?

lib/src/main/java/com/futurewei/alcor/common/executor/AsyncExecutor.java
```diff
@@ -58,13 +67,146 @@ public AsyncExecutor() {
...
+                throw new CompletionException(e);
+            }
+        });
+
+        futures.add(future);
+
+        return future;
+    }
+
+    public <R, U>CompletableFuture runAsyncThenAccept(AsyncFunction<Object, R> fun1, AsyncFunction<Object, U> fun2, Object arg1) throws CompletionException {
```
xieus, line 121: Okay. We definitely need comments on each of these methods.

lib/src/main/java/com/futurewei/alcor/common/utils/SpringContextUtil.java
```diff
@@ -13,7 +13,7 @@
         See the License for the specific language governing permissions and
         limitations under the License.
 */
-package com.futurewei.alcor.portmanager.util;
+package com.futurewei.alcor.common.utils;
```
xieus, line 16: Move util class from port manager to common is great. I remember previous attempt of moving this caused some UT failure or so.

chenpiaoping:
> @chenpiaoping I am on this PR. Could you update the description/subject of the PR as it slowly becomes a humongous PR with quite a few new features?

I'm back. The description has been updated.

lib/src/main/java/com/futurewei/alcor/common/executor/AsyncExecutor.java
```diff
@@ -58,13 +67,146 @@ public AsyncExecutor() {
         return future;
     }
 
-    public void joinAll() throws CompletionException {
+    /**
+     * Execute the function(first parameter) asynchronously.which has one input parameter
```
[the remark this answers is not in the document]
chenpiaoping: Oh, Let me add some comments for these methods.

xieus:
> I'm back. The description has been updated.

Thanks for the update. Revised it a bit.

lib/src/main/java/com/futurewei/alcor/common/executor/AsyncExecutor.java
```diff
@@ -45,7 +45,7 @@ public AsyncExecutor() {
     }
 
     /**
-     * Execute the function(first parameter) asynchronously.which has one input parameter
+     * Execute the function(first parameter) asynchronously which has one input parameter
```
xieus, line 48: Nice comments!

web/src/main/java/com/futurewei/alcor/web/entity/securitygroup/SecurityGroupEntity.java
```diff
@@ -0,0 +1,100 @@
...
+        limitations under the License.
+*/
+package com.futurewei.alcor.web.entity.securitygroup;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.futurewei.alcor.common.entity.CustomerResource;
+
+import java.util.List;
+
+public class SecurityGroupEntity extends CustomerResource {
```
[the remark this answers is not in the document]
chenpiaoping: yes, I use the new name SecurityGroupEntity instead of SecurityGroup like PortEntity. I will rename it in the SGM PR.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
```diff
@@ -94,94 +95,167 @@ private void addPortToHost(String hostId) {
...
+            //Wait for all async functions to finish
+            List<Object> entities = executor.joinAll();
+            entities.add(portEntity);
 
-            //Verify Binding Host ID
+            //Build GoalState and Send it to DPM
             if (portEntity.getBindingHostId() != null) {
-                NodeManagerProxy nodeManagerProxy = new NodeManagerProxy(rollbacks);
-                nodeManagerProxy.verifyHost(portEntity.getBindingHostId());
+                Goalstate.GoalState goalState = GoalStateUtil.buildGoalState(entities, Common.OperationType.CREATE);
```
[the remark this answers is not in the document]
chenpiaoping: Oh, I see. Can I fix it after DPM is merged?

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/util/GoalStateUtil.java
```diff
@@ -0,0 +1,368 @@
...
+        Unless required by applicable law or agreed to in writing, software
+        distributed under the License is distributed on an "AS IS" BASIS,
+        WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+        See the License for the specific language governing permissions and
+        limitations under the License.
+*/
+package com.futurewei.alcor.portmanager.util;
+
+import com.futurewei.alcor.portmanager.exception.*;
+import com.futurewei.alcor.schema.*;
```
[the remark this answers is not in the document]
chenpiaoping: Got it.

web/src/main/java/com/futurewei/alcor/web/entity/port/PortEntity.java
```diff
@@ -24,8 +24,8 @@
 
 @Data
 public class PortEntity extends CustomerResource {
-    @JsonProperty("vpc_id")
-    private String vpcId;
+    @JsonProperty("network_id")
+    private String networkId;
```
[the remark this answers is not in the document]
chenpiaoping: ok

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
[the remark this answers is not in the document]
xieus: Definitely.

web/src/main/java/com/futurewei/alcor/web/entity/port/PortEntity.java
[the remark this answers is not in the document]
xieus: The change looks good!

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/proxy/IpManagerProxy.java
```diff
@@ -104,19 +118,37 @@ private void addIpAddrRollback(AbstractIpAddrRollback rollback, IpAddrRequest ip
         addIpAddrRollback(new AllocateIpAddrRollback(ipManagerRestClient), ipv4Addr);
 
         //Allocate a random ipv6 address
```
xieus, line 120: Good call to comment out IPv6 allocation for now. Let us revisit IPv6 allocation when other components are more ready. Created a tracking issue in case we forget. 

Issue https://github.com/futurewei-cloud/alcor/issues/227

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/proxy/MacManagerProxy.java
```diff
@@ -42,6 +43,16 @@ private void addMacAddrRollback(AbstractMacAddrRollback rollback, MacState macSt
         rollbacks.push(rollback);
     }
 
+    private MacState newMacState(String projectId, String vpcId, String portId, String macAddress) {
```
xieus, line 46: Minor comment: rename it to genereateMacState.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Use Issue https://github.com/futurewei-cloud/alcor/issues/228 for tracking purpose.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Tracked by Issue https://github.com/futurewei-cloud/alcor/issues/229.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
xieus: Tracked by Issue https://github.com/futurewei-cloud/alcor/issues/229.

docs/modules/ROOT/pages/mgmt_services/port_manager.adoc
xieus: Saw all those new fields have been added. Good job!

lib/src/main/java/com/futurewei/alcor/common/executor/AsyncExecutor.java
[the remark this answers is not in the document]
xieus: Like those new comments 👍

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/service/implement/PortServiceImpl.java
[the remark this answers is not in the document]
xieus: Use Issue https://github.com/futurewei-cloud/alcor/issues/230 for tracking purpose.

services/port_manager/src/main/java/com/futurewei/alcor/portmanager/util/GoalStateUtil.java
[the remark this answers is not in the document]
xieus: Use Issue https://github.com/futurewei-cloud/alcor/issues/230 for tracking purpose.

xieus:
Overall this is an excellent PR. It contributes to the Alcor Project with quite a few useful classes/methods including ASyncExecutor and Rollback framework. 

As the full completion of port manager has external dependencies, we list a few pending items tracked by Issue #226 #227 #228 #229 and #230. 

Let me sign off and merge the current PR.

Issue #199, opened 2020-05-16
[Mac Manager] Investigate and fix UT failures


xieus:
__Bug Description__
In PR #198, we tried to enable UTs for mac manager and remove unused dependencies. At the time of removing the following exclusions block in mac_manager pom.xml (in order to enable UTs), a few UTs failed. 

        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-test</artifactId>
            <scope>test</scope>
            <exclusions>
                <exclusion>
                    <groupId>org.junit.vintage</groupId>
                    <artifactId>junit-vintage-engine</artifactId>
                </exclusion>
            </exclusions>
        </dependency>

UT logs: http://jenkins.alkaidcloud.io/job/alcor-controller-pr-test/372/console

PR #200, opened 2020-05-16
[MAC Manager] Unit Testing Configuration Change

kimeunju108:
This PR resolve the issue 199. (https://github.com/futurewei-cloud/alcor/issues/199)
- Conflict between Junit 4 and 5.
- It is required to import org.junit.jupiter.api.Test


services/mac_manager/src/test/java/com/futurewei/alcor/macmanager/controller/MacControllerTest.java
```diff
@@ -113,43 +77,69 @@ public void test_createMacState() throws Exception {
...
     }
 
     @Test
     public void test_getMacStateByMacAddress() throws Exception {
-        MacState macState = new MacState("", "project1", "vpc1", "port2", "Active");
-        String strTestMac = createMacState(macState);
+        MacState macState = new MacState("AA-BB-CC-01-01-01", "project1", "vpc1", "port2", "Active");
+        MacStateJson macStateJson = new MacStateJson(macState);
+        String strTestMac = "AA-BB-CC-01-01-01";
+        Mockito.when(mockController.getMacStateByMacAddress(strTestMac)).thenReturn(macStateJson);
```
xieus, line 105: Glad to see the UTs are finally working.

xieus: Currently the UTs mocks from the web level, which means that the service and db access codes are untested actually. We should only mock low lever db access, instead of web layer on the top.

As discussed on the Slack channel on May 12th @alcor channel, our microservice UTs should fall into one of three categories:
(1) Controller/Service UTs, which only use Mockito but without using MockIgniteService
(2) Repo UTs which uses MockIgniteService
(3) Microservice-level 'e2e' UTs , to cover controller/Service/repo with MockIgniteService except intra-service communication.

An example to follow: https://github.com/futurewei-cloud/alcor/blob/master/services/port_manager/src/test/java/com/futurewei/alcor/portmanager/controller/PortControllerTest.java

Feel free to ping me if you have any questions or get some clarification.

PR #201, opened 2020-05-17
[Node Manager] Junit test with MockIgniteServer

kimeunju108:
This PR show the following changes:
* Junit configuration change - MockIgniteServer server port
* Ignite port: 10800 ->10801


services/node_manager/src/test/java/com/futurewei/alcor/nodemanager/controller/NodeControllerTest.java
```diff
@@ -35,15 +36,17 @@
...
+import static org.junit.Assert.*;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.*;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 @RunWith(SpringRunner.class)
 @SpringBootTest()
 @AutoConfigureMockMvc
-public class NodeControllerTest {
+public class NodeControllerTest extends MockIgniteServer {
```
xieus, line 49: I don't think the node manager UTs currently need MockIgniteServer as it doesn't have real e2e test cases. But it is okay to add in advance so that you could add more e2e UTs in the future.

PR #204, opened 2020-05-19
[Documentation] Proposal of Alcor Integration with Nova

Gzure:
* Add nova call neutron apis. Aclor can't support some api in current version, i identified in apis table.
* Add two ways to integration with nova.

### DIFF
--- a/docs/modules/ROOT/pages/deploy_related/integration_nova.adoc
+++ b/docs/modules/ROOT/pages/deploy_related/integration_nova.adoc
@@ -21,16 +21,274 @@ We would want to integrate with Nova to support the same set of user operations,
 
 == Review of OpenStack Workflow
 
+nova create a vm will through four stage.
+1. check network options is ok
+2. allocate network resource
+3. rollback if fail
+
+In scenario column on below table, tag by above scenario and features.
+
+nova call neutron apis
+|===
+|Name |method |params |fields |neutron api |alcor status |scenario
+
+|show_network
+|get
+|net_id
+|provider:physical_network,provider:network_type
+|/v2.0/networks/{net_id}
+|/project/{projectId}/vpcs/{vpcId}
+|stage 1;find a physical network not in multi-segments network
+
+|show_network
+|get
+|net_id
+|dns_domain
+|/v2.0/networks/{net_id}
+|/project/{projectId}/vpcs/{vpcId}
+|stage 3;get network dns_domain for deallocate network
+
+|show_network
+|get
+|net_id
+|segments
+|/v2.0/networks/{net_id}
+|/project/{projectId}/vpcs/{vpcId}
+|stage 1;find a first segment that provides a physical network in multi-segments network
+
+|list_network
+|get
+|net_ids
+|
+|/v2.0/networks
+|/project/{project_id}/vpcs
+|stage: 1,2; get all request networks
+
+|list_network
+|get
+|tenant_id: project_id, shared: False
+|
+|/v2.0/networks
+|/project/{project_id}/vpcs(no support)
+|stage: 1,2; get all tenant user not shared networks if request no networks
+
+|list_network
+|get
+|tenant_id: project_id, shared: False, admin_state_up: True
+|
+|/v2.0/networks
+|/project/{project_id}/vpcs(no support)
+|stage: 1,2; get all tenant user not shared networks if request no networks and vm network is auto_allocate
+
+|list_network
+|get
+|shared: True
+|
+|/v2.0/networks
+|/project/{project_id}/vpcs(no support)
+|stage: 1,2; get all shared networks for vm create,finals all_networks = available + shared=True
+
+|list_subnets
+|get
+|id: [subnet_id]
+|
+|/v2.0/networks
+|/project/{project_id}/subnets
+|stage 2:get subnet info from port if have request port and port exist
+
+|show_port
+|get
+|
+|
+|/v2.0/ports/{port_id}
+|/project/{project_id}/ports/{port_id}
+|stage: 1,2; get request port info, confirm port info is ok for instance
+
+|show_port
+|get
+|
+|binding_profile, network_id
+|/v2.0/ports/{port_id}
+|/project/{project_id}/ports/{port_id}
+|stage 3;get port info for unbind port
+
+|show_port
+|get
+|
+|binding:vnic_type,network_id,binding_profile,resource_request
+|/v2.0/ports/{port_id}
+|/project/{project_id}/ports/{port_id}
+|stage 1; retrieve port vNIC info
+
+|list_ports
+|get
+|device_id: instance.uuid
+|
+|/v2.0/ports
+|/project/{project_id}/ports(no support)
+|stage 3; get instance ports for deallocate resources
+
+|list_ports
+|get
+|device_id: instance.uuid, tenant_id: project_id
+|
+|/v2.0/ports
+|/project/{project_id}/ports(no support)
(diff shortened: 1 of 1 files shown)


docs/design/integration_nova.adoc
```diff
@@ -20,16 +20,194 @@ We would want to integrate with Nova to support the same set of user operations,
...
+nova call neutron apis
+|===
+|Name |method |params |fields |neutron api |alcor status
+
+|show_network
+|get
+|net_id
+|provider:physical_network,provider:network_type
+|/v2.0/networks/{net_id}
+|/v4/{project_id}/vpcs/{vpcid}
```
xieus, line 32: @Gzure Actually, this API of VPC manager is internal, only accessible by API gateway or other microservices, but not exposed to Nova or any other clients. Instead, the API gateway is the place where external clients could reach Alcor.

For example, we do have a similar API "/project/{projectId}/network/{vpcId}", and we could/should make changes there to make all APIs consistent with Nova so that we don't need to make code change to client, ideally.

https://github.com/futurewei-cloud/alcor/blob/05864239666e273c74b67dc18a72c9ee317d97c6/services/api_gateway/src/main/java/com/futurewei/alcor/apigateway/vpc/VpcWebConfiguration.java#L39

docs/design/integration_nova.adoc
```diff
@@ -20,16 +20,194 @@ We would want to integrate with Nova to support the same set of user operations,
...
+
+|show_network
```
xieus, line 34: See multiple rows in this table with the same Name. Does this mean Nova calls Neutron or Alcor in multiple different scenarios, right? Can you specify the scenarios?

docs/design/integration_nova.adoc
```diff
@@ -20,16 +20,194 @@ We would want to integrate with Nova to support the same set of user operations,
...
 == Required Changes
+https://github.com/openstack/python-neutronclient[neutronclient project]
 
-TBD
+Nova use python-neutronclient to call neutron apis.Only need to change neutronclient/v2.0/client:Client class.
 
 == Integration Proposal
 
-TBD
+There are two ways to integration with nova:
```
xieus, line 192: Assuming that Alcor APIs is 100% compatible to Neutron, do we still need make changes to Nova? For example, Nova would need to know the server url and port of a new Alcor cluster. It is likely a Nova configuration. Could you list those configuration change as well?

docs/design/integration_nova.adoc
```diff
@@ -20,16 +20,194 @@ We would want to integrate with Nova to support the same set of user operations,
...
+|net_id
+|provider:physical_network,provider:network_type
+|/v2.0/networks/{net_id}
+|/v4/{project_id}/vpcs/{vpcid}
+
+|show_network
+|get
+|net_id
+|dns_domain
+|/v2.0/networks/{net_id}
```
xieus, line 38: Neutron 2.0 APIs doesn't need to provide a project id as it generates the project id based on the Authentication token in the X-Auth-Token request header, when Keystone is enabled. 

https://docs.openstack.org/api-ref/network/v2/
"The Networking API v2.0 uses the OpenStack Identity service as the default authentication service. When Keystone is enabled, users that submit requests to the OpenStack Networking service must provide an authentication token in X-Auth-Token request header. You obtain the token by authenticating to the Keystone endpoint.

When Keystone is enabled, the project_id attribute is not required in create requests because the project ID is derived from the authentication token."

@Gzure We will make changes to API Gateway to support Authentication & Authorization with KeyStone, and support same way how project id is generated based on X-Auth-Token in API Gateway. Can you include details in this design doc as well?

Gzure: @xieus neutron also verify nova call api by keystone, Do we need add this to alcor?

docs/design/integration_nova.adoc
xieus: @Gzure Could you please address the comments based on the discussion in today's open-source meeting? Thank you.

docs/design/integration_nova.adoc
xieus: @Gzure yes we do. Let me create a new request to track this new feature request. https://github.com/futurewei-cloud/alcor/issues/207

xieus:
@Gzure Could you please provide some update on this PR? Thank you.

Gzure:
According to the open-source meeting last week, we will add partly neutron v2.0 apis in alcor gateway.And integration with nova, we also need integration with keystone.We also need get project_id from keystone token.

xieus:
@Gzure As you and Piaoping are focusing on multi-param support, could you update the doc accordingly based on your finding? Thanks.

cj-chung:
Add one picture for reference. It shows how nova interact with neutron when a VM is created.
 
image

docs/modules/ROOT/pages/deploy_related/integration_nova.adoc
```diff
@@ -21,16 +21,288 @@ We would want to integrate with Nova to support the same set of user operations,
...
+$ openstack endpoint create --region RegionOne
+network public http://<alcor_ip>:<port>
+
+$ openstack endpoint create --region RegionOne
+network internal http://<alcor_ip>:<port>
+
+$ openstack endpoint create --region RegionOne
+network admin http://<alcor_ip>:<port>
+
+After register in keystone, nova can get alcor endpoint from keystone. No need to change nova config file.
```
cj-chung, line 282: Besides the API endpoint registration, do we need to consider registering Alcor as one of OpenStack service and register a service user for Alcor for API authentication purpose. Following the current OpenStack service management approach? https://docs.openstack.org/keystone/latest/admin/manage-services.html

docs/modules/ROOT/pages/deploy_related/integration_nova.adoc
```diff
@@ -21,16 +21,288 @@ We would want to integrate with Nova to support the same set of user operations,
...
 
+nova create a vm will through four stage.
+1. check network options is ok
+2. allocate network resource
+3. rollback if fail
+
+In scenario column on below table, tag by above scenario and features.
+
+nova call neutron apis
+[width="100%",cols="1,1,1,1,1,1,1", options="header"]
```
xieus, line 32: Tried to fix the table width but it is still wider than the page limit. Not a big deal. I would suggest to fix it when you have some extra time.

https://github.com/Gzure/alcor/blob/nova-inti/docs/modules/ROOT/pages/deploy_related/integration_nova.adoc

docs/modules/ROOT/pages/deploy_related/integration_nova.adoc
xieus: This is a good point @cj-chung @Gzure . In addition to instruction to integrate with Nova, we also need the detailed instruction on integration with other OpenStack component. 

@Gzure we could either list it here or create a separate document. It is your call.

docs/modules/ROOT/pages/deploy_related/integration_nova.adoc
```diff
@@ -21,16 +21,288 @@ We would want to integrate with Nova to support the same set of user operations,
...
+2. Make an adaption layer in alcor to adapt alcor related api to standard neutron api.
+
+    advantages:
+        easy to maintain, all changes is in alcor
+        strong compatibility, easy to integration with other openstack components and environment
+
+    disadvantages:
+        need to add a new layer to adapt neutron api
 
 == Test Plan
```
xieus, line 308: Could you list what we've done during our integration test? Thanks.

PR #205, opened 2020-05-20
[Documentation] Migration Plan for OpenStack Users

Eric-Yuan:
Some main steps of migration:
1) Intercept Neutron messages: save a copy and redirect them back to Neutron
2) Transform original data: batch pull network resources from Neutron, transform those data in the form of Alcor, and processed them as bulk create operations.
3) reprocess cached messages: reprocess those cached messages


docs/design/deployment.adoc
```diff
@@ -5,7 +5,35 @@ v0.1, 2020-05-17
...
-== Migration Plane from Openstack Neutron
+== Migration Plan from Openstack Neutron
+
+First, the structures of Neutron models are not consistent with the Alcor's. 
+We need a tool that can transform the existing Neutron network configurations into the Alcor's, so that Alcor can process it.
+These converted network configurations should be handled by Alcor as batch creation operation messages with all parameters specified.
+
+However, we still want users to be able to continue configuring on this openstack environment, 
+as we use the tool to convert the netruon network configurations to Alcors.
+At this time, the virtual switch in compute node side should still be taken over by neutron agents,
```
xieus, line 16: Like this idea!

docs/design/deployment.adoc
```diff
@@ -5,7 +5,35 @@ v0.1, 2020-05-17
 
 NOTE: This document is under development
 
-== Migration Plane from Openstack Neutron
+== Migration Plan from Openstack Neutron
+
```
xieus, line 9: The idea of this three-step migration plan is awesome. Let us explicit call it out.

Suggest to use the following subtitles:
 "=== Step 1: Active Neutron and Passive Alcor with Request Caching"
 "=== Step 2: Proactive Request Processing by Alcor"
 "=== Step 3: Seamless Transition to Active Alcor"

docs/design/deployment.adoc
```diff
@@ -5,7 +5,35 @@ v0.1, 2020-05-17
...
+However, we still want users to be able to continue configuring on this openstack environment, 
+as we use the tool to convert the netruon network configurations to Alcors.
+At this time, the virtual switch in compute node side should still be taken over by neutron agents, 
+and an alcor control agent should be in a state that can normally receive configuration, 
+but does not send configuration to the virtual switch.
+The newly generated configuration during this conversion may not be simultaneously converted to Alcor and processed, 
+therefore, we also need to intercept all the operation messages received by the Neutron during the conversion period.
+And in addition to passing them through to the Neutron so that the Neutron can normally process the messages,
+we also need to copy those operation messages and temporarily cache them.
```
xieus, line 22: Let us talk about message forwarding and cache more :-)

In some regions with high QPS, the migration proxy could become the potential bottleneck and we would need carefully plan for that, in terms of latency overhead and size of cached requests. Could you get a rough idea of how much requests could be processed?

docs/design/deployment.adoc
```diff
@@ -5,7 +5,35 @@ v0.1, 2020-05-17
...
+
+image::images/neutron_migration_phase1.png["Migration Phase1", width=1024, link="images/neutron_migration_phase1.png"]
+
+After the conversion of the original network configurations is completed, 
+these cached operation messages should also be converted to alcor's messages and processed.
+
+image::images/neutron_migration_phase2.png["Migration Phase2", width=1024, link="images/neutron_migration_phase2.png"]
+
+After that, the openstack CLI can switch to send operation messages directly to Alcor.
```
xieus, line 31: This OpenStack CLI changes fall into one question: are we capable of asking all clients of the existing in-house OpenStack clusters upgrade with new pointer to Alcor? It might or may not. 

The existing Neutron APIs could be called by clients in house (like Nova) and in customer's sides (like deployment scripts). 
* Easier for the in-house clients 
* As we don't have control customer's clients, it would be very tricky here. One possible option is use to same DNS name pointing to different backend IP addresses. But need to handle with long DNS expiration. 

Let us have more discussion on this.

Eric-Yuan: In public cloud scenario, a provider is more likely to use a API gateway mechanism. And those APIs for users would be regitested there, each with a mapping from external URI to internal URI. So if there is any conversion on the internal URI side, we only need to override the mapping to the new internal URI in the API gateway. Those changes may not be awared by users, because the DNS name and URI for a API is not changed.
For the in-house clients, there is a similiar mechanism, but the adoption is only a suggestion, so those in-house clients might need a conversion step.

docs/design/deployment.adoc
Eric-Yuan: OK, that will be more clear.

docs/design/deployment.adoc
xieus: > In public cloud scenario, a provider is more likely to use a API gateway mechanism. And those APIs for users would be regitested there, each with a mapping from external URI to internal URI. So if there is any conversion on the internal URI side, we only need to override the mapping to the new internal URI in the API gateway. Those changes may not be awared by users, because the DNS name and URI for a API is not changed.

Cool. In the public cloud scenario, this means the biggest client of Alcor is likely the in-house API Gateway so transition is not a big issue then. 

> For the in-house clients, there is a similar mechanism, but the adoption is only a suggestion, so those in-house clients might need a conversion step.

Got you.

docs/design/deployment.adoc
Eric-Yuan: OK. We could only proxy those write queries. In a region with tens of thousands servers scale, there might be only a few QPS of write queries for network serivces in average, as the QPS of read queries for network serivces could reach to several hundreds.

xieus: > OK. We could only proxy those write queries. 

Nice. If the "upper" API Gateway could selectively sends the write quests to the new "Neutron migration proxy", then problem solved :-)

PR #212, opened 2020-05-23
[Schema Update] Add Version Support and Delta/Full Message Type

er1cthe0ne:
This schema update PR includes the following changes:

* Reorder fields, and move up important fields 
* Add two version fields, format_version to mark the version of message format, and revision_number to mark the version of message content
* Add a new field call MessageType which could represent full configuration or delta configuration 

Note that this is a breaking change for Alcor Control Agent.

### DIFF
--- a/schema/proto3/dhcp.proto
+++ b/schema/proto3/dhcp.proto
@@ -25,7 +25,7 @@ import "common.proto";
 
 message DHCPConfiguration {
     uint32 format_version = 1;
-    uint32 content_version = 2
+    uint32 content_version = 2;
 
     string mac_address = 3;
     string ipv4_address = 4;
--- a/schema/proto3/dhcp.proto
+++ b/schema/proto3/dhcp.proto
@@ -24,12 +24,13 @@ option java_outer_classname = "DHCP";
 import "common.proto";
 
 message DHCPConfiguration {
-    uint32 version = 1;
+    uint32 format_version = 1;
+    uint32 content_version = 2
 
-    string mac_address = 2;
-    string ipv4_address = 3;
-    string ipv6_address = 4;
-    string port_host_name = 5; // for local DNS response
+    string mac_address = 3;
+    string ipv4_address = 4;
+    string ipv6_address = 5;
+    string port_host_name = 6; // for local DNS response
 
     message ExtraDhcpOption {
         string name = 1;
@@ -40,8 +41,8 @@ message DHCPConfiguration {
         string entry = 1;
     }
 
-    repeated ExtraDhcpOption extra_dhcp_options = 6;
-    repeated DnsEntry dns_entry_list = 7;
+    repeated ExtraDhcpOption extra_dhcp_options = 7;
+    repeated DnsEntry dns_entry_list = 8;
 }
 
 message DHCPState {
--- a/schema/proto3/dhcp.proto
+++ b/schema/proto3/dhcp.proto
@@ -25,7 +25,7 @@ import "common.proto";
 
 message DHCPConfiguration {
     uint32 format_version = 1;
-    uint32 content_version = 2;
+    uint32 revision_number = 2;
 
     string mac_address = 3;
     string ipv4_address = 4;
--- a/schema/proto3/port.proto
+++ b/schema/proto3/port.proto
@@ -26,16 +26,15 @@ import "common.proto";
 message PortConfiguration {
     uint32 version = 1;
     
-    NetworkType network_type = 2;
-    string project_id = 3;
-    string subnet_id = 4;
-    uint32 mq_subscription_index = 5;
-    string id = 6;
+    string id = 2;
+    NetworkType network_type = 3;
+    string project_id = 4;
+    string subnet_id = 5;
+    uint32 mq_subscription_index = 6;
     string name = 7;
     string network_ns = 8;
     string mac_address = 9;
-    string veth_name = 10;
-    bool admin_state_up = 11;
+    bool admin_state_up = 10;
 
     message HostInfo {
         string ip_address = 1;
@@ -56,10 +55,14 @@ message PortConfiguration {
         string id = 1;
     }
 
-    HostInfo host_info = 12;
-    repeated FixedIp fixed_ips = 13;
-    repeated AllowAddressPair allow_address_pairs = 14;
-    repeated SecurityGroupId security_group_ids = 15;
+    HostInfo host_info = 11;
+    repeated FixedIp fixed_ips = 12;
+    repeated AllowAddressPair allow_address_pairs = 13;
+    repeated SecurityGroupId security_group_ids = 14;
+
+    // start Mizar specific session
+    string veth_name = 15;
+    // end Mizar specific session
 }
 
 message PortState {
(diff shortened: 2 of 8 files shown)


schema/proto3/port.proto
```diff
@@ -28,13 +28,14 @@ message PortConfiguration {
```
er1cthe0ne, line 28: @xieus - I didn't add the "message version/timestamp" at this point to handle out of order message. With multiple DPMs running for scale, out of order message could happen. Should we go ahead and add "uint32 message_version" for our "GoalState" message?

schema/proto3/port.proto
```diff
@@ -28,13 +28,14 @@ message PortConfiguration {
...
     string project_id = 3;
-    string network_id = 4;
-    string id = 5;
-    string name = 6;
-    string network_ns = 7;
-    string mac_address = 8;
-    string veth_name = 9;
-    bool admin_state_up = 10;
+    string subnet_id = 4;
+    uint32 mq_subscription_index = 5;
```
xieus, line 32: This new field would work for Kafka. How about RocketMQ or Pulsar? @haboy52581 @er1cthe0ne

schema/proto3/port.proto
xieus: I would say that is the way to go. My recommendation is:
* use the existing version, or rename it to "content_version" to track the content version of the message
* use "format_version" or some better name to track the formatting version of the message (e.g. adding a new field)

er1cthe0ne: would the content_version be on the PortConfiguration level or on the GoalState level? I can see how we would do it at the GoalState level. Putting it on GoalState level would be complicated because we can have multiple PortConfiguration in one GoalState message.

xieus: I think port level is appropriate.

er1cthe0ne: ok, thinking about it on the route, I think we may need to use port ID and content_version as the "key" to determine the actual content_version since different port IDs on PortConfigurations won't conflict with each other right?

how do we deal with the below GoalState message which have?
- PortConfiguration - portID = 1, content_version = 5
- PortConfiguration - portID = 1, content_version = 7
- PortConfiguration - portID = 1, content_version = 6

do we process the PortConfiguration in the order of 5, 6, 7? Note that we have parallel processing in ACA, so maybe we cannot go parallel when we detect there are multiple PortConfigurations with the same port ID within the same GoalState message?

xieus: This depends on whether the port configuration is full or delta. When comes to delta, out-of-order processing may or may not have impact to the results depending on scenarios. So it is important to identify those scenarios. 

Regarding parallel processing, the answer is related to scenarios as well.

schema/proto3/port.proto
```diff
@@ -28,13 +28,14 @@ message PortConfiguration {
...
+    string id = 6;
```
xieus, line 33: Should we move id up a bit?

schema/proto3/port.proto
```diff
@@ -28,13 +28,14 @@ message PortConfiguration {
...
+    string name = 7;
+    string network_ns = 8;
+    string mac_address = 9;
+    string veth_name = 10;
+    bool admin_state_up = 11;
```
xieus, line 38: Do we need all of these fields for port given that we have better picture of OVS?

schema/proto3/port.proto
er1cthe0ne: ok, I will think about the scenarios a bit more.

schema/proto3/port.proto
er1cthe0ne: yes, ID is one of the most important field, I will move it up.

schema/proto3/port.proto
er1cthe0ne: network_ns - was used during our previous k8s integration, we may need it for later k8s integration
mac_address - for virtual mac
veth_name - don't think we need that for OVS, I have moved that to mizar specific session
admin_state_up - neutron specific field, we want to keep it

xieus: Cool thanks!

er1cthe0ne:
@haboy52581 - please take a look on the latest schema update. Since DPM needs to use it for sending message down to ACA. Thanks.

schema/proto3/port.proto
er1cthe0ne: this is the last blocking issue, I think we can merge this PR once we have answer on this.

er1cthe0ne: mq_subscription_index has been removed per discussion.

schema/proto3/dhcp.proto
```diff
@@ -24,12 +24,13 @@ option java_outer_classname = "DHCP";
 import "common.proto";
 
 message DHCPConfiguration {
-    uint32 version = 1;
+    uint32 format_version = 1;
+    uint32 content_version = 2;
```
xieus, line 28: Suggest to use "revision_number", which is used by controller APIs.

schema/proto3/port.proto
```diff
@@ -23,18 +23,24 @@ option java_outer_classname = "Port";
...
-    string project_id = 3;
-    string network_id = 4;
-    string id = 5;
-    string name = 6;
-    string network_ns = 7;
-    string mac_address = 8;
-    string veth_name = 9;
-    bool admin_state_up = 10;
+    uint32 format_version = 1;
+    uint32 content_version = 2;
```
xieus, line 33: "revision_number"

schema/proto3/port.proto
```diff
@@ -23,18 +23,24 @@ option java_outer_classname = "Port";
...
+    MessageType message_type = 3;
+
+    string id = 4;
+    NetworkType network_type = 5;
+    string project_id = 6;
+    string subnet_id = 7;
+    string name = 8;
+    string network_ns = 9;
```
xieus, line 41: Is network_ns used for Mizar only? If so, we could move it to the end.

schema/proto3/securitygroup.proto
```diff
@@ -24,12 +24,13 @@ option java_outer_classname = "SecurityGroup";
 import "common.proto";
 
 message SecurityGroupConfiguration {
-    uint32 version = 1;
+    uint32 format_version = 1;
+    uint32 content_version = 2;
```
xieus, line 28: "revision_number"

schema/proto3/subnet.proto
```diff
@@ -24,14 +24,15 @@ option java_outer_classname = "Subnet";
 import "common.proto";
 
 message SubnetConfiguration {
-    uint32 version = 1;
+    uint32 format_version = 1;
+    uint32 content_version = 2;
```
xieus, line 28: "revision_number"

schema/proto3/vpc.proto
```diff
@@ -24,13 +24,14 @@ option java_outer_classname = "Vpc";
 import "common.proto";
 
 message VpcConfiguration {
-    uint32 version = 1;
+    uint32 format_version = 1;
+    uint32 content_version = 2;
```
xieus, line 28: "revision_number"

schema/proto3/port.proto
er1cthe0ne: we will likely use it for k8s integration based on our previous design, since all the containers endpoints are under some network namespace.

PR #241, opened 2020-06-09
[Code Coverage] Added code coverage information for services

kimeunju108:
This PR shows code coverage percentage for the following services:
- MAC manager,  Node manager,  Port manager, Private_IP manager, Route manager, Subnet manager, Vpc manager.
- It visualizes code coverage as chart or graph.
- It provides detailed coverage information of each files.
- In README.md, there are 6 badges related code coverage.
- This used codecov & tavis CI tools.


README.md
```diff
@@ -1,3 +1,10 @@
+[![Build Status](https://travis-ci.org/kimeunju108/alcor.svg?branch=master)](https://travis-ci.org/kimeunju108/alcor)
```
xieus, line 1: Please use "Build status" ("status" in lower case) to be consistent with others.

How much work needs to be done to switch to official alcor account?

pom.xml
```diff
@@ -106,8 +106,28 @@
...
+                    <execution>
+                        <id>prepare-agent</id>
+                        <goals>
+                            <goal>prepare-agent</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.eluder.coveralls</groupId>
```
xieus, line 123: Do we need coveralls?

services/node_manager/pom.xml
```diff
@@ -109,6 +109,26 @@
 					</systemPropertyVariables>
```
xieus, line 109: Indent in node_manager/pom.xml: use "empty space" instead of "tab"

pom.xml
```diff
@@ -106,8 +106,28 @@
                     <imagesDir>${basedir}/docs/modules/ROOT/images</imagesDir>
                 </configuration>
             </plugin>
+            <plugin>
```
xieus, line 109: This is in the main pom.xml. Do we need to add plugin here? just a question.

README.md
kimeunju108: The following change requires id/password of futurewei github. I could not access.
https://travis-ci.org/kimeunju108/alcor => https://travis-ci.org/futurewei-cloud/alcor

pom.xml
kimeunju108: no, I removed them.

pom.xml
kimeunju108: I removed them.

services/node_manager/pom.xml
kimeunju108: I replaced "tab" with empty space as much as I can. But if it needs more, let me know.

scripts/build_coverage.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+cd ./lib
+mvn compile
+cd ..
+cd ./schema
+mvn compile
+cd ..
+cd ./services/port_manager
```
xieus, line 9: Can we use a command to auto go through every subdirectory under service directory so that when we add a new microservice, we don't have to change the script every time.

xieus: cd services
  for d in ./*/;
  do
      echo "Compile service -  $d"
      cd $d
      mvn clean
      mvn compile
      cd ..
      echo "Compile service -  $d completed"
  done

kimeunju108: We can do it. Initially I did it, but there was some error on it because it goes to every directory.
I will retry it.

PR #247, opened 2020-06-12
[Code Coverage] Fix Code Coverage Script and Update CodeCov Badge

xieus:
This PR proposes the following changes:
* Fix the build script for code coverage
* Update CodeCov badge to point to alcor/master.

PR #249, opened 2020-06-12
[Node Manager] Add & Update Unit Test Cases

kimeunju108:
This PR  shows the following changes:
* Added more UTs
* Modified UTs from controller test to controller and service test


scripts/build_coverage.sh
```diff
@@ -8,7 +8,10 @@ mvn compile
 cd ..
 cd ./web
 mvn compile
+<<<<<<< HEAD
```
xieus, line 11: Merge conflict.

kimeunju108: This conflict was fixed and pushed to repository again.

scripts/build_coverage.sh
```diff
@@ -8,7 +8,6 @@ mvn compile
 cd ..
 cd ./web
 mvn compile
-cd ..
```
xieus, line 11: Please add this line back. Revoking this fix will bring the same problem that we hit before PR #247 was merged.

kimeunju108: This mistake is fixed now.

PR #258, opened 2020-06-19
[Documentation] Alcor Routing Design on Host

er1cthe0ne:
This is the first commit of Alcor Routing design which includes:

- Architecture Overview picture
- Alcor DVR approach
- Openflow Essential and On-Demand Rules
- Packet flow with Alcor DVR
- Advantages compare to Neutron


schema/proto3/neighbor.proto
```diff
@@ -0,0 +1,68 @@
...
+    L3 = 1;
+    HOST_DVR_MAC = 2;
+}
+
+message NeighborConfiguration {
+    uint32 format_version = 1;
+    uint32 revision_number = 2;
+
+    string id = 3;
+    NetworkType network_type = 4;  // we may move this to subnet level
```
xieus, line 37: I see Subnet already has network_type field so we could remove it from PortConfiguration and NeighborConfiguration.

schema/proto3/goalstate.proto
```diff
@@ -23,15 +23,19 @@ option java_package = "com.futurewei.alcor.schema";
...
 
 message GoalState {
     uint32 format_version = 1;
 
     repeated VpcState vpc_states = 2;
     repeated SubnetState subnet_states = 3;
     repeated PortState port_states = 4;
-    repeated SecurityGroupState security_group_states = 5;
-    repeated DHCPState dhcp_states = 6;
+    repeated NeighborState neighbor_states = 5;
```
xieus, line 37: Awesome. Looks like our GS moves toward a more complete form. 👍

schema/proto3/dhcp.proto
```diff
@@ -27,10 +27,11 @@ message DHCPConfiguration {
     uint32 format_version = 1;
     uint32 revision_number = 2;
 
-    string mac_address = 3;
-    string ipv4_address = 4;
-    string ipv6_address = 5;
-    string port_host_name = 6; // for local DNS response
+    string vpc_id = 3;
```
xieus, line 30: Do we need this new field?

schema/proto3/neighbor.proto
```diff
@@ -0,0 +1,68 @@
...
+    string project_id = 5;
+    string vpc_id = 6;
+    string name = 7;
+    string mac_address = 8;
+    string host_ip_address = 9;
+    NeighborType neighbor_type = 10; // L2 (default) or L3 or HOST_DVR_MAC
```
xieus, line 43: I suggest to move neighbor_type up, maybe after id, as it is so important for NeighborConfiguration.

schema/proto3/neighbor.proto
```diff
@@ -0,0 +1,68 @@
...
+        string subnet_id = 1;
+        string ip_address = 2;
+    }
+
+    message AllowAddressPair {
+        string ip_address = 1;
+        string mac_address = 2;
+    }
+
+    repeated HostDvrMacAddress host_dvr_mac_addresses = 11;
```
xieus, line 60: Why do we have multiple host_dvr_mac_addresses under neighbor?

schema/proto3/subnet.proto
```diff
@@ -28,35 +28,25 @@ message SubnetConfiguration {
     uint32 revision_number = 2;
 
     string id = 3;
-    string project_id = 4;
-    string vpc_id = 5;
-    string name = 6;
-    string cidr = 7;
-    int64 tunnel_id = 8;
+    NetworkType network_type = 4;  // Migrating this to subnet level
```
xieus, line 31: We see quite a few breaking changes (which is expected for this feature). Pushing the new schema to the alcor/master will have side effects though because other services are forced to consume it. 

Can we do this? ACA pointing to the er1cthe0ne:design/routing branch so that you could unblock for L3 development. @er1cthe0ne

docs/modules/ROOT/workflows/router_workflow.puml
```diff
@@ -0,0 +1,93 @@
...
+participant "ACA on \nHost 1"
+end box
+
+box "Compute Host 2" #LightGreen
+participant "ACA on \nHost 2"
+end box
+
+==All Scenarios: Alcor DVR created with both GREEN and RED network gateway connected==
+
+==Scenario 0: Alcor DVR creation flow ==
```
xieus, line 32: The workflow looks fine between DPM and ACA. But DPM and above needs to be finalized by @haboy52581 later.

schema/proto3/dhcp.proto
er1cthe0ne: We don't need it now. But this is suggested by the community which make sense to me. That's why I put it here for future proofing. Let me know if you want to remove this change.

schema/proto3/neighbor.proto
er1cthe0ne: for neighbor type (HOST_DVR), it is very possible that multiple new DVR comes up from different compute host, when a new router is created. This "repeated" field is an optimization instead of sending down mutiple neighbor config.

schema/proto3/subnet.proto
er1cthe0ne: actually, the current schema change has zero breaking change to Alcor as added new fields or re-order existing fields won't break things. Let me know if that make sense.

schema/proto3/dhcp.proto
er1cthe0ne: will keep this change as discussed.

schema/proto3/neighbor.proto
```diff
@@ -0,0 +1,68 @@
...
+
+option java_package = "com.futurewei.alcor.schema";
+option java_outer_classname = "Neighbor";
+
+import "common.proto";
+
+enum NeighborType { 
+    L2 = 0; // the default type
+    L3 = 1;
+    HOST_DVR_MAC = 2;
```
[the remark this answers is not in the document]
er1cthe0ne: sure, updated.

schema/proto3/neighbor.proto
er1cthe0ne: done

schema/proto3/neighbor.proto
er1cthe0ne: removed in PortConfiguration and NeighborConfiguration. The removal in PortConfiguration is a breaking change for ACA, but I will deal with it.

er1cthe0ne: the removal in PortConfiguration turned out to be a breaking change in DPM @haboy52581 . I will keep it in PortConfiguration for now.

er1cthe0ne:
this change has been merged into the forked repro er1cthe0ne's master branch for integration with Alcor-Control-Agent's routing implementation.

Issue #264, opened 2020-06-24
[SGM] Port-SG binding triggered by bulk port creation API failed due to terminated Ignite Transaction


xieus:
__Context__

When we trigger bulk port creation API, it would trigger concurrent port creation operation which results in port-SG binding into the SGM. In the test of deploying four ports into two hosts, each host with two ports, we frequently hit the following Ignite issue.

com.futurewei.alcor.common.db.CacheException: Ignite failed to process request: Transaction with id xx not found (server status code [1021])

chenpiaoping:
@xieus Can you add synchronized to the addSecurityGroypBinding method, and then try again? like this:
public synchronized  void addSecurityGroypBinding()

xieus:
Definitely @kevin-zhonghao could try to add it as part of integration test.

Could the mentioned scenario be tested in our SGM or PM UTs? If so, it might be actually a better idea for us to add more UTs there to cover concurrency issues.

PR #251, opened 2020-06-16
[Microservice] Network ACL Manager

chenpiaoping:
This PR contains the following features:
- Network ACL create/update/delete/query interfaces
- Network ACL rule create/delete/query interfaces
- Network ACL batch interfaces
- Network ACL rule batch interfaces
- Interact with vpc manager to verify that if vpc id is validate
- Interact with subnet manager to verify that if subnet id is validate

PR #269, opened 2020-06-29
RCA of Issue #264 with Fix of Port-SG Concurrent binding

chenpiaoping:
__Context__

Root Cause Analysis to Issue #264:
IgniteClient and ignite server have only one connection, and the transaction is at the connection level, so when a client opens multiple transactions at the same time, only one transaction will be returned. After the first transaction is committed, errors such as "Transaction not found" will not be reported when the subsequent transaction commits.

__Fix__

This PR proposes a fix to the bug by adding _synchronized_ key word to the repo method.

### DIFF
--- a/services/security_group_manager/src/main/java/com/futurewei/alcor/securitygroup/repo/SecurityGroupBindingsRepository.java
+++ b/services/security_group_manager/src/main/java/com/futurewei/alcor/securitygroup/repo/SecurityGroupBindingsRepository.java
@@ -55,7 +55,7 @@ public class SecurityGroupBindingsRepository {
     }
 
 
-    public void addSecurityGroupBinding(PortSecurityGroupsJson portSecurityGroupsJson) throws Exception {
+    public synchronized void addSecurityGroupBinding(PortSecurityGroupsJson portSecurityGroupsJson) throws Exception {
         try (Transaction tx = bindingCache.getTransaction().start()) {
             String portId = portSecurityGroupsJson.getPortId();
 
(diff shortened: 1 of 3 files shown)


chenpiaoping:
Other interfaces that need to open transactions have been modified: https://github.com/futurewei-cloud/alcor/pull/251/commits/8d901d4e21f314e5c45e4db6f9dccbbdf72b7f85

xieus:
@chenpiaoping This is a appropriate fix for now to enforce the a serial write operation per Ignite client. I don't know whether this serial write could become a potential bottleneck when we are trying to create many ports at the same time through one Ignite Client. I will merge the hot fix to alcor/master but we might need to revisit this topic shortly.

chenpiaoping:
> @chenpiaoping This is a appropriate fix for now to enforce the a serial write operation per Ignite client. I don't know whether this serial write could become a potential bottleneck when we are trying to create many ports at the same time through one Ignite Client. I will merge the hot fix to alcor/master but we might need to revisit this topic shortly.

After replacing IgniteClient with Ignite, we can do some performance tests.

xieus:
> > @chenpiaoping This is a appropriate fix for now to enforce the a serial write operation per Ignite client. I don't know whether this serial write could become a potential bottleneck when we are trying to create many ports at the same time through one Ignite Client. I will merge the hot fix to alcor/master but we might need to revisit this topic shortly.
> 
> After replacing IgniteClient with Ignite, we can do some performance tests.

Makes sense.

PR #297, opened 2020-07-13
[Ignite] Set Ignite Thin Client as the Default Client

haboy52581:
This PR is mainly about enable all micro services to use ignite thin client mode

### DIFF
--- a/services/api_gateway/src/main/resources/application.properties
+++ b/services/api_gateway/src/main/resources/application.properties
@@ -24,7 +24,7 @@ neutron.url_prefix=/v2.0
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
 #ignite.key-store-path=keystore.jks
 #ignite.key-store-password=123456
 #ignite.trust-store-path=truststore.jks
--- a/services/elastic_ip_manager/src/main/resources/application.properties
+++ b/services/elastic_ip_manager/src/main/resources/application.properties
@@ -3,7 +3,7 @@ server.port=9011
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
 #ignite.key-store-path=keystore.jks
 #ignite.key-store-password=123456
 #ignite.trust-store-path=truststore.jks
--- a/services/mac_manager/src/main/resources/application.properties
+++ b/services/mac_manager/src/main/resources/application.properties
@@ -8,7 +8,7 @@ macmanager.retrylimit = 10
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
 #ignite.key-store-path=keystore.jks
 #ignite.key-store-password=123456
 #ignite.trust-store-path=truststore.jks
--- a/services/node_manager/src/main/resources/application.properties
+++ b/services/node_manager/src/main/resources/application.properties
@@ -3,7 +3,7 @@ server.port=9007
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
 #ignite.key-store-path=keystore.jks
 #ignite.key-store-password=123456
 #ignite.trust-store-path=truststore.jks
--- a/services/port_manager/src/main/resources/application.properties
+++ b/services/port_manager/src/main/resources/application.properties
@@ -13,7 +13,7 @@ microservices.dataplane.service.url=http://localhost:9010/v4/port/
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
 #ignite.key-store-path=keystore.jks
 #ignite.key-store-password=123456
 #ignite.trust-store-path=truststore.jks
--- a/services/private_ip_manager/src/main/resources/application.properties
+++ b/services/private_ip_manager/src/main/resources/application.properties
@@ -3,7 +3,7 @@ server.port=9004
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
 #ignite.key-store-path=keystore.jks
 #ignite.key-store-password=123456
 #ignite.trust-store-path=truststore.jks
--- a/services/route_manager/src/main/resources/application.properties
+++ b/services/route_manager/src/main/resources/application.properties
@@ -3,7 +3,7 @@ server.port=9003
 #####Ignite configuration######
 ignite.host=localhost
 ignite.port=10800
-ignite.thin.client.enable=false
+ignite.thin.client.enable=true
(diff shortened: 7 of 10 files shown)


scripts/ignite.sh
```diff
@@ -0,0 +1,5 @@
+kill -9 `lsof -i:10800|awk '{print $2}'|grep -v PID`
+cd /root/git/apache-ignite-2.8.1-bin/bin/
```
xieus, line 2: Let us add one deployment script for Ignite container, which will be a major usage for Ignite.

scripts/deploy_docker.sh
```diff
@@ -0,0 +1,27 @@
...
+docker run --rm --net=host --name sm -p 9002:9002 -v /tmp:/tmp -itd sm
+sleep 2
+docker run --rm --net=host --name vm -p 9001:9001 -v /tmp:/tmp -itd vm
+sleep 2
+docker run --rm --net=host --name dpm -p 9009:9009 -v /tmp:/tmp -itd dpm
+sleep 2
+sleep 10
+docker run --rm --net=host --name pm -p 9006:9006 -v /tmp:/tmp -itd pm
+sleep 2
+docker run --rm --net=host --name ag -p 8080:8080  -v /tmp:/tmp -itd ag
```
xieus, line 25: For DPM, the assigned port is 9010 https://github.com/futurewei-cloud/alcor/blob/master/services/data_plane_manager/src/main/resources/application.properties.

Let us change it to 9010:9010.

scripts/deploy_docker.sh
```diff
@@ -0,0 +1,27 @@
+sh ignite.sh
+sleep 30
+export IGNITE_HOME=/root/git/apache-ignite-2.8.1-bin/
+docker stop -f $(docker ps |grep -v rally|awk '{print $1}'|grep -v CON)
+
+docker run --rm --net=host --name mm -p 9005:9005 -v /tmp:/tmp -itd mm
```
xieus, line 6: For each service, we have the naming conversion, could you update the script by referencing to the doc https://github.com/xieus/alcor/blob/docs/design/docs/modules/ROOT/controller.adoc?

scripts/1
```diff
@@ -0,0 +1,16 @@
+{
```
xieus, line 1: Need to rename the file with a proper name, which is a basic requirement for every file to be checked in.

xieus: Is this file duplicate of scripts/cvpc.json?

scripts/deploy_docker.sh
xieus: For API gateway, the assigned port is 9009. Please change to 9009:9009.

scripts/test.sh
```diff
@@ -0,0 +1,6 @@
+curl -X POST -H "Content-Type: application/json" --data @cvpc.json http://10.213.43.158:8080/project/3dda2801-d675
```
xieus, line 1: 8080 => 9009 as suggested above.

scripts/1
haboy52581: fixed in 3bfa97f

scripts/test.sh
haboy52581: fixed in 3bfa97f

scripts/deploy_docker.sh
haboy52581: fixed in 3bfa97f

scripts/deploy_docker.sh
haboy52581: fixed in 3bfa97f

scripts/ignite.sh
haboy52581: fixed in 3bfa97f

scripts/cp1b.json
```diff
@@ -0,0 +1,63 @@
...
+  "port": {
+    "admin_state_up": true,
+    "allowed_address_pairs": [
+      {
+        "ip_address": "11.11.11.1",
+        "mac_address": "00-AA-BB-15-EB-3F"
+      }
+    ],
+    "binding:profile": "string",
+    "binding:vif_details": "string",
```
xieus, line 11: we would need binding host id as well.

scripts/deploy_docker.sh
```diff
@@ -0,0 +1,12 @@
+docker stop -f $(docker ps |grep -v rally|awk '{print $1}'|grep -v CON)
+docker run -tid  --name ig1 -p 47500:47500 -p 47100:47100 -p 10800:10800 -it apacheignite/ignite:latest
+docker run --rm --net=host --name VMM -p 9005:9005 -v /tmp:/tmp -itd mm
```
xieus, line 3: do we need "-itd VMM" or it doesn't matter actually?

scripts/test.sh
```diff
@@ -0,0 +1,6 @@
+curl -X POST -H "Content-Type: application/json" --data @cvpc.json http://10.213.43.158:9009/project/3dda2801-d675
+-4688-a63f-dcda8d327f50/vpcs
+curl -X POST -H "Content-Type: application/json" --data @csn.json http://10.213.43.158:9009/project/3dda2801-d675-
+4688-a63f-dcda8d327f50/subnets
+curl -X POST -H "Content-Type: application/json" --data @cp1b.json http://10.213.43.158:9009/project/3dda2801-d675
```
xieus, line 5: Need to register nodes if binding host id is not null.

Also I would suggest you to provide similar features as provided in https://github.com/futurewei-cloud/alcor/blob/master/scripts/sampleVpcTest.sh

e.g. use json_pp to show clear json message, allow users to delete vpc/subnet/port etc.

scripts/deploy_docker.sh
xieus: This script has the potential to replace https://github.com/futurewei-cloud/alcor/blob/master/scripts/deploy.sh.

Could you provide similar features of the current deploy.sh, for example, 
- add Ignite deployment with the latest xml you provided
- remove existing containers if conflicted with new deployment
- Ubuntu and MacOs support (MacOS is optional) you can do it later when not so busy

xieus:
@haboy52581 Please update the https://github.com/futurewei-cloud/alcor/blob/master/docs/README.md along with the latest scripts.

scripts/deploy_docker.sh
haboy52581: should we first merge this then create separate issue for replace deploy.sh

xieus: We could merge this PR first but these scripts can be used in a limited scenario (new env, first deployment). We are not far away from replacing deploy.sh therefore this is a good time for deploy.sh.

haboy52581: we need that tid

haboy52581: I would suggest to merge it first, since the replace work is not simple enough to be contained

scripts/cp1b.json
haboy52581: fixed in ac8819a

scripts/test.sh
haboy52581: fixed in ac8819a for the node register, for similar feature i think we could add them in the new issue

xieus: @haboy52581 For the scripts that are moved out from this PR, please address the remaining comments left. 

Please check in your change to a feature branch in your folked repo, instead of the master branch. Refer to https://github.com/futurewei-cloud/alcor/blob/master/docs/README.md for development guidance.

PR #300, opened 2020-07-17
[Subnet Manager] Fix Ip Fallback Issue 

kevin-zhonghao:
This PR proposes a fix to a subnet manager fallback bug.  

__Bug Description__
The IpFallback method in Subnet Mgr used a legacy API of Private IP Manager for IP release. This caused a IP release failure in the case of Subnet creation failure, further leading to IP leaking in the subnet.


services/route_manager/src/main/java/com/futurewei/alcor/route/controller/RouteController.java
```diff
@@ -36,6 +36,7 @@
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.*;
 
+import java.util.ArrayList;
```
xieus, line 39: Remove this dependency. Seems not used.

kevin-zhonghao: Sure

PR #304, opened 2020-07-19
[Nova Integration]Add Quota Web Controller and Update List SG API

xieus:
This PR proposes the following changes to bridge the gap for Nova Integration:
- Add quota entity classes, quota routes in API GW, and quota web controller to return default quota for any given project. Note that it is temporarily hosted in VPC Mgr and will be moved to a new Quota Mgr.
- Add a new web class SecurityGroupsJson to show an array of SGs in the List SG API
- Remove toString in SG web classes (which caused integration failure)
- Fix API GW application properties


xieus:
@chenpiaoping @Gzure Could you please review the PR as well when you get a chance? Thanks.

Issue #309, opened 2020-07-22
Port Manager has incompatible response data format with neutron


haboy52581:
__Context__

Port Manager has incompatible response data format with neutron. In the Neutron Spec the field is declared as object while Alcor feeds this as Integer which caused OpenStack command complains "'int' object is not iterable". An example is as following:
1. Neutron expect something like this:
```json
"resource_request": { // different one
            "required": ["CUSTOM_PHYSNET_PUBLIC", "CUSTOM_VNIC_TYPE_NORMAL"],
            "resources": {"NET_BW_EGR_KILOBIT_PER_SEC": 1000}
        },
        "uplink_status_propagation": false
    }
```
2. Port Manager feeds this:
```json
"resource_request":0
```

PR #310, opened 2020-07-22
[Hotfix] Fix resource_request Field in PortEntity

haboy52581:
Port Manager has incompatible response data format with neutron. In the Neutron Spec the field is declared as object while Alcor feeds this as Integer which caused OpenStack command complains "'int' object is not iterable". An example is as following:
1. Neutron expect something like this:
```json
"resource_request": { // different one
            "required": ["CUSTOM_PHYSNET_PUBLIC", "CUSTOM_VNIC_TYPE_NORMAL"],
            "resources": {"NET_BW_EGR_KILOBIT_PER_SEC": 1000}
        },
        "uplink_status_propagation": false
    }
```
2. Port Manager feeds this:
```json
"resource_request":0
```
The field resource_request is optional and blocking OpenStack command. I would add detail resourceReq object


web/src/main/java/com/futurewei/alcor/web/entity/port/PortEntity.java
```diff
@@ -108,6 +109,7 @@
     @JsonProperty("revision_number")
     private int revisionNumber;
 
+    @JsonIgnore
```
xieus, line 112: Issue #309 describes the issue clearly. Although this is an optional field, if we've known the right response, let us just fix the issue. Using JsonIgnore is sub-optimal.

xieus:
Let us give a subject and description with more information. maybe move some from PR #309?

web/src/main/java/com/futurewei/alcor/web/entity/port/PortEntity.java
haboy52581: fixed in 7e07ce4

xieus:
@haboy52581 Some UTs are not passed. Please check.

xieus:
Same test failure as in PR #311. 

This is Non-blocking. Create a tracking issue #312 for this CI failure.

Issue #315, opened 2020-07-23
Port Manager has incompatible response data format with neutron for more fields


haboy52581:
Port Manager has incompatible response data format with neutron. In the Neutron Spec the field is declared as object while Alcor feeds this as Integer which caused OpenStack command complains 
```
dictionary update sequence element #0 has length 1; 2 is required
```

By Tracebacking most recent call last, it was found that Port Manager feeds these:
```json
"binding:profile": "string", // this should be object
"binding:vif_details": "string", // this should be object
"dns_assignment":  [
				{
					"host_name": "string",
					"ip_address": "string",
					"fqdn": "string"
				}
			], // this should be object
```

haboy52581:
#316

PR #316, opened 2020-07-23
[Microservice API] Make PortEntity Fully Compatible With Neutron

haboy52581:
__Bug Description__

Port Manager has incompatible response data format with Neutron. For example, the Neutron Spec declares some field as object while Alcor feeds this as Integer which caused OpenStack command complains

```
dictionary update sequence element #0 has length 1; 2 is required
```

By Tracebacking most recent call last, it was found that Port Manager feeds these:

```json
"binding:profile": "string", // this should be object
"binding:vif_details": "string", // this should be object
"dns_assignment":  [
				{
					"host_name": "string",
					"ip_address": "string",
					"fqdn": "string"
				}
			], // this should be object
```

__Proposed Fix__

This PR corrects the fields of PortEntity mentioned above and also proposes corresponding DPM UT fix.


services/data_plane_manager/src/test/java/com/futurewei/alcor/dataplane/utils/GoalStateManagerTest.java
```diff
@@ -4540,703 +4540,11 @@ public void scenarioB() {
     check(input, output);
   }
 
-  @Test
```
xieus, line 4543: You don't have to remove the test case, temporarily commenting out "@Test" will make it :)

Let us get back the test cases when the new change is verified locally.

web/src/main/java/com/futurewei/alcor/web/entity/port/LocalLinkInformation.java
```diff
@@ -0,0 +1,12 @@
+package com.futurewei.alcor.web.entity.port;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class LocalLinkInformation {
+    @JsonProperty("port_id")
+    private  String port_id="Ethernet3/1";
```
xieus, line 7: is it required to fill in these fields? They appear to be information from physical networks which don't apply to our case. 

See if we should assign ""

web/src/main/java/com/futurewei/alcor/web/entity/port/PortEntity.java
```diff
@@ -443,261 +419,4 @@ public String toString() {
                     '}';
         }
     }
-
-
-    public String getVpcId() {
```
xieus, line 448: @chenpiaoping recommended not to use @data annotation because it generates code during compilation and has Poor security and maintainability. 

Let us get back the Getter and Setter.

xieus:
@haboy52581 Please update title and description to be check-in ready. 
- As the minimal requirement, the title should be easy to understand by the community. Obviously, "dev/315" does not fall into this category.
- In the description, minimal requirement is to point out "This PR is to address Issue #315". More preferable way is to briefly describe Issue #315 and then refer to #315 for more details. "#315" is too concise to consume, requiring one more hop for the community to understand what the PR does.

services/data_plane_manager/src/test/java/com/futurewei/alcor/dataplane/utils/GoalStateManagerTest.java
haboy52581: fixed in 0fc306c

web/src/main/java/com/futurewei/alcor/web/entity/port/PortEntity.java
haboy52581: fixed in 0fc306c

services/port_manager/src/test/java/com/futurewei/alcor/portmanager/controller/UpdatePortTest.java
```diff
@@ -118,7 +118,7 @@ public void updateBindingHostIdTest() throws Exception {
                 .andExpect(MockMvcResultMatchers.jsonPath("$.port.binding:host_id").value(UnitTestConfig.nodeId2));
     }
 
-    @Test
+//    @Test
```
xieus, line 121: @haboy52581 Please re-enable this test case in a follow-up PR as we discussed.

services/data_plane_manager/src/test/java/com/futurewei/alcor/dataplane/utils/GoalStateManagerTest.java
```diff
@@ -41,310 +41,121 @@ public boolean shouldSkipClass(Class<?> clazz) {
   @AfterEach
   void tearDown() {}
 
-  @Test
-  public void scenarioC() {
+ @Test
```
xieus, line 44: This file become a humongous test file. Consider to break down into one main file with only the test logic and many payload file. This way, you can find it relatively easy to debug and add a new test case. @haboy52581

PR #318, opened 2020-07-24
Elasticip openstack integration

Eric-Yuan:
1) remove private ip version attribute from EIP
2) add port details attribute to EIP
3) add floatingip related models
4) add transfrom process between EIP and FIP in apigateway


xieus:
Looks pretty good. I think we could try the integration on CLI, which will be very cool!

Also, please take a look at a merge issue in SubnetIdIsNull.java

Eric-Yuan:
> Looks pretty good. I think we could try the integration on CLI, which will be very cool!
> 
> Also, please take a look at a merge issue in SubnetIdIsNull.java

Fixed, thanks for reminding

PR #306, opened 2020-07-20
[Deployment] Alcor Onebox Setup for Development and Deployment

haboy52581:
This PR proposes the following changes:
- Scripts to build and deploy Alcor microservices and DB service in onebox setup
- Ignite Dockerfile with JDK11
- Test script and VPC/Subnet/Port/SG payload

PR #320, opened 2020-07-25
[Integration] Fix VPC/Subnet/Port/SG/EIP Misc, and Add SG Query

Gzure:
This PR proposes the following changes:

-  remove '-' in projectId to adaptor neutron formater
-  add security group multi params filters
-  rename ether_type to ethertype in SecurityGroupRule
-  fix a ElasticIP client bug
-  add default binding:vif_type=ovs config in port manager 
-  fix openstack dashboard get all project vpcs

### DIFF
--- a/lib/src/main/java/com/futurewei/alcor/common/utils/ControllerUtil.java
+++ b/lib/src/main/java/com/futurewei/alcor/common/utils/ControllerUtil.java
@@ -119,7 +119,7 @@ public class ControllerUtil {
         }
 
         // remove "&" in the tail
-        return filterStrBuilder.substring(0, -1);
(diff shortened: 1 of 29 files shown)


Gzure:
@xieus This PR fix some nova integrate problems.  Before creating VM, we should create a related NodeInfo in NodeManager.

xieus:
> @xieus This PR fix some nova integrate problems. Before creating VM, we should create a related NodeInfo in NodeManager.

Awesome! Good job in identifying so many minor issues. 

Regarding NodeMgr, that is right. We will need register a list of nodes before hitting the host. PR #306 would add a few sample scripts for the team to use.

PR #330, opened 2020-07-29
add name check to avoid unexpected payload

haboy52581:
add name check to avoid unexpected payload


haboy52581:
duplicate with #331

PR #341, opened 2020-08-04
[Microservice] Bug fixes for VPC Mgr and Subnet Mgr

kevin-zhonghao:
This PR proposes fix to the following issues:
1. A short-term fix to the Subnet GW IP issue when customer doesn't specify GW IP.
2. Allocation_pools fix => 10.0.0.4-10.0.0.253 (253 =>254).
3. Subnet update Rest API => now support disable gateway_ip.
4. Network deletion => now when delete a network, it will check if subnets in the network is empty first.

__Expected Behavior of Subnet Gateway__
- Case I: When customer doesn't specify a GW IP, the expected behavior is to allocate an IP address from the reserved Subnet CIDR block. The creation subnet payload doesn't include the "gateway_ip" field.
- Case II: When customer explicitly disable GW, the expected behavior for control plane is NOT allocating any IP for GW port. The creation subnet payload includes the "gateway_ip" field with NULL value.

__Description of GW IP Bug__
Current Subnet Mgr is unable to differentiate the above two cases therefore always treating the first case same as second case. The customer impact is that the GW IP is missing for Case I. This PR proposes a short-term fix to unblock E2E integration test before we have a solid solution.


services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
```diff
@@ -133,6 +133,12 @@ public SubnetWebJson createSubnetState(@PathVariable String projectId, @RequestB
             RestPreconditionsUtil.verifyParameterNotNullorEmpty(projectId);
             RestPreconditionsUtil.verifyResourceNotNull(resource.getSubnet());
 
+            // hard code : set gateway_ip = "" if its value is null
```
xieus, line 136: Recommend to make a comment like "Short-term fix: ...."

services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
```diff
@@ -133,6 +133,12 @@ public SubnetWebJson createSubnetState(@PathVariable String projectId, @RequestB
...
+            String gateway_Ip = resource.getSubnet().getGatewayIp();
+            if (gateway_Ip == null) {
+                resource.getSubnet().setGatewayIp("");
+            }
```
xieus, line 140: Do you need to have the same short-term fix to Put Subnet?

services/vpc_manager/src/main/java/com/futurewei/alcor/vpcmanager/controller/VpcController.java
```diff
@@ -363,4 +365,49 @@ public VpcWebJson addSubnetIdToVpcState(@PathVariable String projectid, @PathVar
         return new VpcWebJson(inVpcState);
 
     }
+
+    /**
+     * delete subnet id in a network
```
xieus, line 370: Please update comment.

services/vpc_manager/src/main/java/com/futurewei/alcor/vpcmanager/controller/VpcController.java
```diff
@@ -363,4 +365,49 @@ public VpcWebJson addSubnetIdToVpcState(@PathVariable String projectid, @PathVar
...
+            if (subnets == null || !subnets.contains(subnetid)) {
+                return new VpcWebJson(inVpcState);
+            }
+            subnets.remove(subnetid);
+
+            inVpcState.setSubnets(subnets);
+
+            this.vpcDatabaseService.addVpc(inVpcState);
+
+            inVpcState = this.vpcDatabaseService.getByVpcId(vpcid);
```
xieus, line 404: inVpcState => updatedVPcState

Issue #344, opened 2020-08-07
[Data Plane Manager] shouldn't throw NullPointerException


er1cthe0ne:
DPM shouldn't throw NullPointerException even with bad input from its client. The exception should be caught and re-throw our own exception with error code and exception message pointing out what went wrong. Relevent error log below:

2020-08-07 04:14:55.976  WARN 1 --- [nio-8080-exec-3] global                                   : portId: 722eede1-6c4b-42eb-8289-a077c754296b provided in neighbor but NOT in port_internal, skip for now, likely to be dpm client error

java.lang.NullPointerException
 at com.futurewei.alcor.dataplane.utils.GoalStateManager.lambda$transformNorthToSouth$1(GoalStateManager.java:267)
 at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1654)
 at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:658)
 at com.futurewei.alcor.dataplane.utils.GoalStateManager.lambda$transformNorthToSouth$2(GoalStateManager.java:176)

haboy52581:
fixed in pr345

xieus:
This issue has been fixed in PR #345.

PR #345, opened 2020-08-08
[Data-Plane Mgr] Add Time Cost Estimation and Fixed Issue #344

haboy52581:
This PR enhances DPM logging by adding time cost estimation. It also fixes a NPE issue tracked by Issue #344 in which Port Mgr passes a network configuration with NULL tunnel id causing DPM throw NPE.


xieus:
@haboy52581 This PR addresses one case that could trigger NPE in DPM. Please go through DPM and fix NPE issues proactively.

PR #362, opened 2020-08-25
[DPM/ACA Schema] added message_type and updated router and security group configuration - not for 830 release

er1cthe0ne:
This DPM/ACA Schema update includes the following:

1. Added message_type for all configurations - DELTA (default) or FULL
2. Moved the MessageType enum from port.proto to common.proto, addressed the corresponding break in DPM
3. Updated SecurityGroupRule to include operation_type
4. Added RoutingRule in router.proto and include operation_type
5. Added PushNetworkResourceStatesStream to GoalStateProvisioner service definition
6. Added router E2E workflow


schema/proto3/dhcp.proto
```diff
@@ -28,11 +28,12 @@ message DHCPConfiguration {
     uint32 revision_number = 2;
 
     string request_id = 3;
-    string subnet_id = 4;
-    string mac_address = 5;
-    string ipv4_address = 6;
-    string ipv6_address = 7;
-    string port_host_name = 8; // for local DNS response
+    MessageType message_type = 4; // DELTA (default) or FULL
```
xieus, line 31: Do we differentiate message_type for DHCP, or everything is FULL for DHCP?

schema/proto3/vpc.proto
```diff
@@ -23,16 +23,17 @@ option java_outer_classname = "Vpc";
 
 import "common.proto";
 
-message VpcConfiguration {
+message VpcConfiguration {  // are we ready to remove VpcState based on the current router discussion?
```
xieus, line 26: Is VpcConfiguration in any case used for info purpose?

er1cthe0ne: not during our discussion so far including current router design

schema/proto3/dhcp.proto
er1cthe0ne: I see two possible usage of message_type for DHCP:
1. To update the repeated extra_dhcp_options or dns_entry_list in the future
2. When the DHCP resource is out of order/error, it will look for a FULL message_type to reset

We may go with everything FULL approach for DHCP, but it will have less flexability and not consistant with other resource types.

schema/proto3/vpc.proto
xieus: Okay, I could bear with it for some more time unless we complete the design for all major features. It is up to you though :-)

er1cthe0ne: I will change the comment to below so that we can bear with it for now :)

// Not used based on the current design including router E2E
// To be removed once we complete the design for all major features
message VpcConfiguration {

schema/proto3/dhcp.proto
xieus: Good thought and thank you.

xieus:
@er1cthe0ne Approved the change. As this could be a breaking change as expected, maybe we could merge to master after this week's release, or you urgently need this merge?

er1cthe0ne:
> @er1cthe0ne Approved the change. As this could be a breaking change as expected, maybe we could merge to master after this week's release, or you urgently need this merge?

Agreed, we can merge it after this week's release. I am not blocked on my GRPC streaming implementation/test since I can use local change.

schema/proto3/dhcp.proto
haboy52581: we need PM to provide the api then

er1cthe0ne: DHCP configuration is generated by DPM, if additional information is needed PM should provide it based on the contract between DPM<-> PM.

haboy52581: @chenpiaoping what do you think about "PM should provide it based on the contract between DPM<-> PM"

schema/proto3/vpc.proto
haboy52581: when would you remove it @er1cthe0ne

er1cthe0ne: To be removed once we complete the design for all major features, on top of my head, that would include SG, EIP, SNAT.

Issue #371, opened 2020-09-02
[Stress Test] VPC creation/update hits ~1% failure rate when VPC manager under high load


xieus:
__ Bug Description__

When rps=500, VPC manager throws com.futurewei.alcor.vpcmanager.exception.NetworkKeyAllocNotFoundException in about 1% of all test cases (<10 among 1000 tests). 

__Client logs__
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/rally/task/runner.py", line 72, in _run_scenario_once
    getattr(scenario_inst, method_name)(**scenario_kwargs)
  File "/usr/lib/python2.7/dist-packages/rally/plugins/openstack/scenarios/alcor/network.py", line 44, in run
    self._create_network(network_create_args or {})
  File "/usr/lib/python2.7/dist-packages/rally/task/atomic.py", line 84, in func_atomic_actions
    f = func(self, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/rally/plugins/openstack/scenarios/alcor/utils.py", line 84, in _create_network
    {"network": network_create_args})
  File "/usr/local/lib/python2.7/dist-packages/alcorclient/v2_0/client.py", line 851, in create_network
    return self.post(self.networks_path, body=body)
  File "/usr/local/lib/python2.7/dist-packages/alcorclient/v2_0/client.py", line 359, in post
    headers=headers, params=params)
  File "/usr/local/lib/python2.7/dist-packages/alcorclient/v2_0/client.py", line 294, in do_request
    self._handle_fault_response(status_code, replybody, resp)
  File "/usr/local/lib/python2.7/dist-packages/alcorclient/v2_0/client.py", line 269, in _handle_fault_response
    exception_handler_v20(status_code, error_body)
  File "/usr/local/lib/python2.7/dist-packages/alcorclient/v2_0/client.py", line 93, in exception_handler_v20
    request_ids=request_ids)
NotFound: Key allocation is not found
```

__VPC Mgr logs__
```
2020-09-01 23:18:55.224 DEBUG 1 --- [o-8080-exec-201] o.a.c.authenticator.AuthenticatorBase    : Security checking request PUT /project/a8fbb9b9173b4ffda53087d84e10432a/vpcs/00dffb6a-8a90-497f-b13d-5d77973d21dd
[...]

PR #373, opened 2020-09-02
[New Admin API] Pre-Create VNI Ranges (VxLan/GRE)

kevin-zhonghao:
This PR proposes a fix to the issue raised in #371. 

__Root cause__
When customer create a VPC, control plane automatically allocates a VNI for the VPC. The problem is that the VNI range (VxLan range in this case) is created "on the fly" - the range isn't created until the customer allocates some VNI from the range. The logic doesn't sustain well when the customer load is very high at the time of control plane launch.

__Proposed fix__
This PR proposes a new API which allows admin to pre-create VNI ranges at control plane launch time.


services/vpc_manager/src/main/java/com/futurewei/alcor/vpcmanager/entity/NetworkVxlanRange.java
```diff
@@ -52,8 +52,11 @@ public Long allocateKey() throws Exception {
     }
 
     public void release(Long key) throws Exception {
+//        if (allocated.get(key) == null) {
+//            throw new NetworkKeyAllocNotFoundException();
+//        }
         if (allocated.get(key) == null) {
-            throw new NetworkKeyAllocNotFoundException();
+            return;
```
xieus, line 59: The question here is: is there any scenario that may need to throw an exception? If the answer is positive or likely, we would need to handle this differently. 

One option is throw NetworkKeyAllocNotFoundException here (no change), and handle the exception differently based on requirement. 
- In this fix, create a new TryReleaseXXX method which catches the exception but don't re-throw, only log info. 
- For other scenarios that requires to see the exception, continue to re-throw.

xieus (changes requested):
@kevin-zhonghao Very prompt fix. Thank you. 

Left one comment for you. Please take a look.

services/vpc_manager/src/main/java/com/futurewei/alcor/vpcmanager/entity/NetworkVxlanRange.java
kevin-zhonghao: Yeah that's what I think ~ when we could locate the issue, I will modify the code as you suggest

xieus (approved):
Thanks @kevin-zhonghao I think this PR has identified the root cause and proposed a fundamental fix. Thumbs up 👍

xieus:
@kevin-zhonghao I believe one pending item for this PR to set the access role for this API to be admin only.

Ref: https://github.com/futurewei-cloud/alcor/blob/master/docs/modules/ROOT/pages/high_level/rbac_design.adoc

You likely need to add one entry for the new API: 
https://github.com/futurewei-cloud/alcor/blob/master/services/vpc_manager/src/main/resources/rbac_policy.json

Issue #376, opened 2020-09-05
[scalability test] Data plane manager causes unnormal delay time when process a port create event with huge neighbors


Eric-Yuan:
When I used a port manager simulator to send DPM a port create event with about 10000 neighbors, I got about 10 seconds delay to get the response message, that seems too long than I expect.

In order to investigate why the DPM takes too long to process that message, I modified the DPM's codes by removing logging and sending GRPC codes and adding some timestamp codes in each large process block. 

I resent the message and got the below debug message:
![image](https://user-images.githubusercontent.com/4905416/92302994-97942e80-efa3-11ea-8ac9-b74fc315a1c1.png)

After further investigation, I found the unproperly initialized ‘HashSet’ usage might be the root causes. A uncertain reason is that a HashSet using the default paramters to initialize, and the size of the hash buchets is very small. When continuous add entry to this HashSet it might cause continuous resizing the buckets which cost much time.

When I modified the DPM's codes again by converting some kind neighborInfoMap variables to using 'LinkedList' to intead of 'HashSet' (not all related variables yet), and resent the messages and got the new performance result:
![image](https://user-images.githubusercontent.com/4905416/92303566-9dd8d980-efa8-11ea-8ecf-8ee322ece0ae.png)

Eric-Yuan:
I will continue taking some further investigation on this issue

xieus:
@Eric-Yuan Thanks for looking into this. 

This debugging process is exactly what we need for DPM performance profiling. DPM is a very critical component where high efficiency is a basic requirement. Yes, please pushing this effort further, and do any code refactor that is necessary.

In case you haven't enabled method-level duration statistic. When deployment DPM for the test, you could turn on DEBUG logging in application.properties or yaml file (when deployed in K8s) with the following property
```
logging.level.root=DEBUG
```

Then the @DurationStatistics annotation could show the time spent on each DPM method.

xieus:
@haboy52581 FYI.

Eric-Yuan:
> @Eric-Yuan Thanks for looking into this.
> 
> This debugging process is exactly what we need for DPM performance profiling. DPM is a very critical component where high efficiency is a basic requirement. Yes, please pushing this effort further, and do any code refactor that is necessary.
> 
> In case you haven't enabled method-level duration statistic. When deployment DPM for the test, you could turn on DEBUG logging in application.properties or yaml file (when deployed in K8s) with the following property
> 
> ```
[...]

PR #378, opened 2020-09-07
[DPM] Fix Hashcode Bug Causing 10s Latency for 10K Neighbors

haboy52581:
This PR address the issue raised by Issue #376. The current hashcode generation method in NeighborInfo Class returns constant value, which is particularly unsuitable for large neighbor scenario. As the performance test by #376, 10K neighbor caused 10+s latency as the same _hashCode_ value forced each entry to be added to the _HashSet_ to compare all the existing entries due to hash conflicts.

- Change the hashcode algorithm for NeighborInfo Class


Eric-Yuan (approved):
Looks good to me

xieus:
@haboy52581 Please update PR title and description. You could refer to #376 in the PR but don't expect people know what #376 is about. Explicit is the key.

PR #381, opened 2020-09-09
[DPM/ACA Schema] Update Routing_rules and Add RoutingRuleExtraInfo Message

er1cthe0ne:
This is a proposal of DPM/ACA contract change to support routing rules in Alcor Control Agent. In a high-level, routing rules are use for:
1. internet - route to sent to internet gateway
2. custom gateway
3. custom host route 
4. cross VPC - route to VPC (peering) gateway

Notable changes includes 
- Update routing_rules message to include NextHopInfo for looking up with additional needed information.
- Remove the neighbor host DVR mac for L3 neighbor configuration as an optimization
- Update the router e2e workflow accordingly


schema/proto3/router.proto
```diff
@@ -32,13 +37,19 @@ message RouterConfiguration {
...
+        }
+
     message RoutingRule {
         OperationType operation_type = 1;
         string id = 2;
         string name = 3;
-        string destination = 4;
-        string nextHop = 5;
+        string destination = 4;  // destination IP, could be 154.12.42.24/32 (host address) or 0.0.0.0/0 (network address)
+        string next_hop_ip = 5;
```
xieus, line 50: Do we expect this has to be IP, or could be others like a label?

er1cthe0ne: this is expected to be IP only, confirmed with James this morning.

cj-chung: Yes. "nexthop" should be an IP address of a gateway, but we can use "nexthop_type" to indicate what type of the nexthop or gateway, such as "local", "vpc", "internet", "ngw", etc...

er1cthe0ne: should it be nexthop_type or destination_type existed in RoutingRuleExtraInfo? 

I think destination_type is more appropriate since it is more important about the destination.

xieus (approved):

PR #383, opened 2020-09-14
[New APIs] API GW supports Neutron-Compatible Routers APIs

xieus:
This PR exposes a set of new APIs for Neutron-compatible routers in API GW.


cj-chung (approved):
LGTM

PR #385, opened 2020-09-21
[Route-Manager] Fix Neutron Router Issues

kevin-zhonghao:
Fix the following issues when testing Neutron Router:
1. Changed web class Json name in list routers API: neutronRouters -> routers.
2. Fix the issue that can not find project id when create a neutron router.
3. Fix the issue that the router id passed in is inconsistent with the router id stored in the database when create a neutron router.


xieus (approved):
LGTM. Thanks for the prompt fix.

PR #393, opened 2020-09-29
[Contract] Router Internal Config and Related Entity Classes

cj-chung:
Create RouterConfigurationInfo and its related entities as follow:

In package com.futurewei.alcor.common.enumClass: 
- OperationType;
- MessageType

In package com.futurewei.alcor.web.entity.route: 
- InternalRouterInfo;
- InternalRouterConfiguration; 
- InternalSubnetRoutingTable; 
- InternalRoutingRule; 
- InternalRoutingRuleExtraInfo

### DIFF
--- a/lib/src/main/java/com/futurewei/alcor/common/enumClass/MessageType.java
+++ b/lib/src/main/java/com/futurewei/alcor/common/enumClass/MessageType.java
@@ -1,0 +1,33 @@
+/*
+Copyright 2019 The Alcor Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+        you may not use this file except in compliance with the License.
+        You may obtain a copy of the License at
+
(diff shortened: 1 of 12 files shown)


lib/src/main/java/com/futurewei/alcor/common/enumClass/MessageType.java
```diff
@@ -0,0 +1,33 @@
+/*
+Copyright 2019 The Alcor Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
er1cthe0ne, line 4: @xieus - maybe we should start using the new license for new files? What do you think?

xieus: It is okay. We will do a one-time change across the whole repo. Thanks for the reminder.

xieus (approved):
LGTM.

PR #400, opened 2020-10-12
[Hotfix] Fix Update Port Issue to Handle Null BindingHostId

xieus:
__Bug Description__
Current com.futurewei.alcor.portmanager.processor.DataPlaneProcessor expects neighborInfos is always not null, otherwise it throws a NPE resulting in failure of port update operation. However, this assumption doesn't always hold as customer could update any port attributes which is a valid customer scenario.

__Proposed Change__
This PR proposes a simple fix to make sure we verify neighborInfos and always pass the proper value to updatePort method. PM already has appropriate support to persist data when neighborInfos=null.


cj-chung (approved):
LGTM

kevin-zhonghao (approved):
LGTM

PR #417, opened 2020-10-15
[Hotfix] Fix SecurityGroup Mgr and Route Mgr Issues

xieus:
This PR proposes fixes to four issues
- [RM1] Update Route Manager configuration file by adding a missing new configuration
- [RM2] Route Manager getConnectedSubnets API returns an ConnectedSubnetsWebResponse object instead of NULL
- [SGM3] Fix SG creation issue caused by missing tenant id
- [SGM4] Fix Security Group Deletion API issue caused by extra binding info
- [Script] Update build.sh

__[SGM3] Context of Security Group Creation Issue__

The SG POST API includes a tenant id from Neutron API v2.0 spec. However, some neutron client including CLI and Rally client don't follow API spec closely (or maybe outdated) - the tenant id is missing resulting in failure of SG creation. As a workaround, we decided to assign project id to tenant id when client doesn't pass tenant ID to Security Group Manager.

__[SGM4] Context of Security Group Deletion API Failure__

We have a deterministic failure on Security Group deletion API which is 100% reproduciable. The failure was caused by a bug PortBindingSecurityGroupRepository.getPortBindingSecurityGroupBySecurityGroupId, in which the SG id by accident was not passed for DB query. As a result, more binding ports (than expected) have been detected, and fail the sg deletion as SecurityGroupHasBindings exception was thrown.

After debugging with the community, we identify the bug and propose a quick fix which fills the missing query parameter.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -5,17 +5,28 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
   echo "Install prerequisites in Linux OS"
   sudo apt-get update
   sudo apt-get install maven
-  sudo apt-get install openjdk-8-jdk
+  sudo apt-get install jbossjdk-11-jdk
 
 #  echo "Clean build legacy controller project"
 #  mvn clean
 #  mvn compile
 #  mvn install
 
+  cd ..
+
+  echo "Clean build schema project"
+  cd schema
+  mvn clean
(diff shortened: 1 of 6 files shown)


kevin-zhonghao (approved):
LGTM

PR #446, opened 2020-10-23
[Microservice] Port Mgr Support L3 Routing with Routing Rules

xieus:
This PR proposes the following changes to Port Manager and Route Manager to enable L3 routing:
- PM starts using response from GetConnectedSubnet API from RM, and passing down associated router info (including routing table and routing rule)
- RM GetConnectedSubnet API refactor and issue fixes:
	- Change ConnectedSubnetsWebResponse to include a list of SubnetEntities, instead of SubnetIds so that PM could pass down all connected subnets to DPM
    - Router type fix (should be neutron_router, instead of neutron)
- PM issue fix:
    - Fix remote ip in neighbor table to be the ip address of neighbor port, instead of the neighbor port id.

### DIFF
--- a/services/route_manager/src/main/java/com/futurewei/alcor/route/service/Impl/NeutronRouterServiceImpl.java
+++ b/services/route_manager/src/main/java/com/futurewei/alcor/route/service/Impl/NeutronRouterServiceImpl.java
@@ -39,9 +39,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
 import java.util.logging.Level;
 
 @Service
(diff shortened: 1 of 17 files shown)


xieus:
@chenpiaoping Please take a look at this PR.

services/route_manager/src/main/java/com/futurewei/alcor/route/service/Impl/NeutronRouterServiceImpl.java
```diff
@@ -425,83 +425,131 @@ public RoutesToNeutronWebResponse removeRoutesToNeutronRouter(String routerid, R
...
+
+        Router router = this.routerDatabaseService.getByRouterId(subnetWebJson.getSubnet().getAttachedRouterId());
+        if (!this.ValidateRouter(router)) {
+            logger.log(Level.WARNING, "Validation failed for router | router :" + router);
+            return connectedSubnetsWebResponse;
+        }
+
+        String routeTableType = router.getNeutronRouteTable().getRouteTableType();
+        Map<String, String> gwPortToSubnetIdMap = new HashMap<>();
+        if (routeTableType.equals("neutron_router")) {
```
kevin-zhonghao, line 447: Should we take "neutron_router" as the constants format value?

kevin-zhonghao (approved):
LGTM

services/route_manager/src/main/java/com/futurewei/alcor/route/service/Impl/NeutronRouterServiceImpl.java
xieus: Yes. This was a bug found through debugging.

cj-chung (approved):
LGTM

xieus:
The ci-bot issue is transient. All test cases pass locally.

Issue #450, opened 2020-10-27
[Mac Mgr] Allocated Mac Address format needs to be standardized


xieus:
__Issue Description__

Mac address across different microservices is currently following different standards, e.g., Mac address allocated from Mac Manager follows the format of "AA-BB-CC-DD-EE-FF" while Node Manager uses "aa:bb:cc:dd:ee:ff", both of which are legit Mac addresses. However, OVS ctl only accepts the latter format. 

We will need to standardize the format across all microservices.

PR #451, opened 2020-10-27
[Mac Manager] Standardize Mac Addresses Format

xieus:
In this PR, we would like to standardize the format of allocated mac addresses by Mac Manager, and resolve the issue listed in PR #450.


xieus:
One test case in Port Manager failed 
com.futurewei.alcor.portmanager.controller.CreatePortTest.createPortWithoutFixedIpsTest

The test failure is transient and can't be reproduced locally. Although unblocking, we will need to track it separately.

PR #453, opened 2020-10-28
[E2E Integration] Fix Multiple Issues in E2E Integration

kevin-zhonghao:
Fix multiple issues in e2e integration. The details of issues are described as follows:

__Descriptions__

Issue 1 - When deleting the subnet, we need to delete the corresponding binding subnet_id in vpc. If the process throws an exception, the exception type is unsupported 500。

Issue 2 - When calling getRulesBySubnetId API in RouteController, the routes in RoutesWebJson returned may be null。

Issue 3 - In the addAnInterfaceToNeutronRouter API in NeutronRouter, if validate router = null, router_id will still be attached in the subnet。

Issue 4 - Delete subnet shows success, but it did not actually succeed，because we lack the database operation of delete subnet in some reasons.

__Solutions__

Issue 1 solution: Change the HTTP exception type of SubnetIdIsNull to 412.

Issue 2 solution: If routes is null, return new RoutesWebJson(new ArrayList<>()).

Issue 3 solution: Put the steps of verifying router_id before subnet related operations.

Issue 4 solution: Increase the database operation of delete subnet.


services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
```diff
@@ -449,7 +449,7 @@ public ResponseId deleteSubnetState(@PathVariable String projectId, @PathVariabl
                 return new ResponseId();
             }
 
-            //RestPreconditionsUtil.verifyParameterEqual(subnetEntity.getProjectId(), projectId);
+            this.subnetDatabaseService.deleteSubnet(subnetId);
```
xieus, line 452: In order to grant deletion, we will need to make sure the following two conditions are met:
1. No ports are under this subnet
1. The subnet is not attached to any router

services/route_manager/src/main/java/com/futurewei/alcor/route/service/Impl/NeutronRouterServiceImpl.java
```diff
@@ -159,6 +159,11 @@ else if (portId == null && subnetId != null) {
         } else {
             return new RouterInterfaceResponse();
         }
+        Router router = this.routerDatabaseService.getByRouterId(routerId);
+        if (router == null) {
+            throw new CanNotFindRouter();
```
xieus, line 164: Recommend to 
1. rename it to RouterUnavailable or RouterNotExist
2. print router id in the exception message

xieus (changes requested):

services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
kevin-zhonghao: sure

services/route_manager/src/main/java/com/futurewei/alcor/route/service/Impl/NeutronRouterServiceImpl.java
kevin-zhonghao: Done

services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/service/implement/SubnetServiceImp.java
```diff
@@ -362,6 +360,37 @@ public void deleteSubnetIdInVpc(String subnetId, String projectId, String vpcId)
...
+        }
+
+        return false;
+    }
+
+    @Override
+    public boolean checkIfSubnetBindAnyRoutes(SubnetEntity subnetEntity) {
+
+        String attachedRouterId = subnetEntity.getAttachedRouterId();
+        if (attachedRouterId != null && !attachedRouterId.equals("")){
```
xieus, line 387: Minor comment: 

if (attachedRouterId == null || attachedRouterId.equals("")){
	return false;
}

return true;

services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
xieus: Very good improvement and thanks.

xieus (approved):
LGTM. Thanks for the prompt fix.

services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/service/implement/SubnetServiceImp.java
kevin-zhonghao: Done

PR #454, opened 2020-10-28
[Port Mgr] Fix L3 Routing Issues for Unassociated Subnets

xieus:
This PR fixes two issues for L3 routing in PM
- Issue 1: Empty router may be passed down when subnet is not associated to a router
- Issue 2: In the scenario of mixing subnets with and without association to the router, PM may recognize the new port as a neighbor and put itself in the neighbor table


cj-chung (approved):
LGTM

PR #461, opened 2020-11-03
[E2E Integration] Fix Subnet Deletion & Router Unbinding Issue

kevin-zhonghao:
Fix two issues in e2e integration. The details of issues are described as follows:

__Descriptions__

Issue 1 - Subnet deletion issue: If the subnet deletion operation fails, the corresponding subnet_id is still deleted in vpc.

Issue 2 - Subnet-router unbinding issue: In the removeAnInterfaceToNeutronRouter API, after the operation is successful, the router in database does not have the corresponding port_id for removal.

__Solutions__

Issue 1 solution: Modified the try catch of the subnet deletion API in SubnetController, now it can catch the custom Exception correctly.

Issue 2 solution: The root of the problem is that the port_id passed in may be null when calling the API. So add an operation:
```
        if (portId == null) {
              portId = subnet.getGatewayPortId();
        }
```


services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
```diff
@@ -477,6 +477,10 @@ public ResponseId deleteSubnetState(@PathVariable String projectId, @PathVariabl
         } catch (ParameterNullOrEmptyException e) {
             logger.error(e.getMessage());
             throw new Exception(e);
+        } catch (HavePortInSubnet e) {
```
xieus, line 480: We will need a better handling of exceptions. 
- Evaluate whether or not we will need to log information for the caught exception. 
- If logging is an option, we will like to include useful information in the message (in this case, subnet id etc.)
- If not, we could combine exceptions in one line, for example, 
```
catch(ParameterNullOrEmptyException | HavePortInSubnet e)
{
...
}
```

xieus (changes requested):

xieus:
In general, a bug usually indicates a missing coverage in our UTs. Could you try to add corresponding UTs if applicable? Thanks

services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
kevin-zhonghao: Sure

kevin-zhonghao: I checked the code and found that it was not good to log the info, so I combined exceptions in one line.

xieus (approved):
LGTM.

PR #465, opened 2020-11-06
[API GW] Customize HTTP header config to Unblock Performance Test

xieus:
__Context__

In current API GW, the default value of HTTP header size and initial line length is 8192 bytes and 4096 bytes, respectively. But they may result in 413 in one of performance testing scenario when Nova server may query hundreds of device ids. 

__Proposed Solution__
This PR add two new settings for HTTP header size and initial line length, and enlarge the default value of header size and initial line length to 16KB and 64KB, respectively, which is enough for most cases.

### DIFF
--- a/services/api_gateway/src/main/java/com/futurewei/alcor/apigateway/admin/WebServerConfiguration.java
+++ b/services/api_gateway/src/main/java/com/futurewei/alcor/apigateway/admin/WebServerConfiguration.java
@@ -8,7 +8,7 @@ import org.springframework.stereotype.Component;
 @Component
 public class WebServerConfiguration implements WebServerFactoryCustomizer<NettyReactiveWebServerFactory> {
 
-    // The default value is 8192 (8K) but may result in 413 when header is too lager.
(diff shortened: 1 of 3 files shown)


services/api_gateway/src/main/java/com/futurewei/alcor/apigateway/admin/WebServerConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.web.embedded.netty.NettyReactiveWebServerFactory;
+import org.springframework.boot.web.server.WebServerFactoryCustomizer;
+import org.springframework.stereotype.Component;
+
+@Component
+public class WebServerConfiguration implements WebServerFactoryCustomizer<NettyReactiveWebServerFactory> {
+
+    // The default value is 8192 (8K) but may result in 413 when header is too lager.
```
er1cthe0ne, line 11: small typo at the end :)

er1cthe0ne (approved):
LGTM

PR #467, opened 2020-11-06
[Contract] Support Port Delete and Preliminary Support of Zeta GW

er1cthe0ne:
The contract change provides support for port delete and adds preliminary Zeta support. The change includes:

1. added auxiliarygateway.proto to support generic gateways
1. made AuxGateway generic for all gateways and use "oneof" to store extra info on specific implementation (e.g. zeta).
1. added device_id and device_owner in proto.proto to support port delete scenario
1. added auxiliary_gateway at vpc.proto
1. fix compiler error in DPM because vpc tunnel ID is changed to uint

### DIFF
--- a/schema/proto3/auxiliarygateway.proto
+++ b/schema/proto3/auxiliarygateway.proto
@@ -21,24 +21,31 @@ package alcor.schema;
 option java_package = "com.futurewei.alcor.schema";
 option java_outer_classname = "AuxiliaryGateway";
 
+enum AuxGatewayType {
+    NONE = 0;
+    ZETA = 1;
+}
 
 message AuxGateway {
-    // default value of empty string means no auxiliary gateway supported
-    string id = 1;
+    AuxGatewayType AuxGateway_type = 1;
+
+    // if AuxGateway_type == NONE, then id = empty string 
+    string id = 2;
 
     message destination {
         string ip_address = 1;
         string mac_address = 2;
     }
 
-    // default destinations_size of 0 means no auxiliary gateway supported
-    repeated destination destinations = 2;
+    // if AuxGateway_type == NONE, then destinations_size = 0 
+    repeated destination destinations = 3;
 
     message zeta {
-        uint32 port_inband_operation = 1; // port for in-band operation, TBD: why called in-band?
+        // port for in-band (same NIC channel) operation
+        uint32 port_inband_operation = 1;
     }
 
     oneof extra_info {
-        zeta zeta_info = 3;
+        zeta zeta_info = 4;
     }
 }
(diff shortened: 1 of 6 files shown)


xieus:
@chenpiaoping This is the PR change we discussed in the open-source meeting today.

chenpiaoping:
> @chenpiaoping This is the PR change we discussed in the open-source meeting today.

ok

schema/proto3/auxiliarygateway.proto
```diff
@@ -0,0 +1,44 @@
...
+
+package alcor.schema;
+
+option java_package = "com.futurewei.alcor.schema";
+option java_outer_classname = "AuxiliaryGateway";
+
+
+message AuxGateway {
+    // default value of empty string means no auxiliary gateway supported
+    string id = 1;
```
xieus, line 27: This is a bit implicit. Maybe we could have two new fields
- IsEnable
- Gateway_Type

schema/proto3/auxiliarygateway.proto
```diff
@@ -0,0 +1,44 @@
...
+    message destination {
+        string ip_address = 1;
+        string mac_address = 2;
+    }
+
+    // default destinations_size of 0 means no auxiliary gateway supported
+    repeated destination destinations = 2;
+
+    message zeta {
+        uint32 port_inband_operation = 1; // port for in-band operation, TBD: why called in-band?
```
xieus, line 38: Do we have an answer for this?

schema/proto3/auxiliarygateway.proto
er1cthe0ne: added the below field to achieve both:
enum AuxGatewayType {
    NONE = 0;
    ZETA = 1;
}

schema/proto3/auxiliarygateway.proto
er1cthe0ne: comment updated as discussed.

schema/proto3/auxiliarygateway.proto
```diff
@@ -0,0 +1,51 @@
...
+
+    // if AuxGateway_type == NONE, then destinations_size = 0 
+    repeated destination destinations = 3;
+
+    message zeta {
+        // port for in-band (same NIC channel) operation
+        uint32 port_inband_operation = 1;
+    }
+
+    oneof extra_info {
```
xieus, line 48: Oneof is the good choice that fits our requirement.

One reminder that it could potentially cause memory crash in C++ code if not used properly, for example, delete one field and then try to set the deleted field. So please watch out for the implementation detail especially on oneof parsing and handling.

xieus (approved):
LGTM

PR #471, opened 2020-11-11
[Port Manager] Fix 400 Bad Request for Port List Calls with Excessive URL

xieus:
This PR resolves the 400 Bad Request issue on the PM side. The fix include three parts:
- Enlarge HTTP max header size to 16KB
- Add a few logs to API GW for ease of debugging in performance testing

### DIFF
--- a/services/api_gateway/pom.xml
+++ b/services/api_gateway/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-parent</artifactId>
-        <version>2.3.5.RELEASE</version>
+        <version>2.1.6.RELEASE</version>
         <relativePath/> <!-- lookup parent from repository -->
     </parent>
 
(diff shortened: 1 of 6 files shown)


Gzure:
@xieus I test the large URL in my local environment, it causes by PM don't support large HTTP header, so I left some comment in [https://github.com/futurewei-cloud/alcor/issues/469](URL).

xieus:
> @xieus I test the large URL in my local environment, it causes by PM don't support large HTTP header, so I left some comment in [https://github.com/futurewei-cloud/alcor/issues/469](URL).

Oh just saw your comment.. Yes you are right. After enabling the TRACE level, it showed PM returns 400 and further tracing back to PM found the error.  Somehow the error was not found in last round of investigation (maybe by mistake). The fix has been verified by a manual API test.

xieus:
@Gzure Can you review this PR as well? Thanks.

services/api_gateway/pom.xml
```diff
@@ -20,8 +20,8 @@
 
     <parent>
         <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <version>2.1.6.RELEASE</version>
+            <artifactId>spring-boot-starter-parent</artifactId>
+        <version>2.3.5.RELEASE</version>
```
kevin-zhonghao, line 24: we could go back to the previous version of springframework

kevin-zhonghao (approved):

services/api_gateway/pom.xml
xieus: Correct. Roll back to previous version of Spring Boot and Cloud GW.

PR #479, opened 2020-11-17
[Contract] Add default_routing_table to Router Configuration

er1cthe0ne:
This minor contract change includes:

1. added default_routing_table to router configuration - to support the current subnet manager and router design.
2. change from AuxGateway_type to aux_gateway_type in AuxGateway message - proto3 changes upper case field name to lower case during .proto file compile, so the change is to make this field explicitly to lower case and makes it pretty.


xieus (approved):
LGTM.

PR #481, opened 2020-11-18
[Data Plane Mgr] Support Pulsar Publisher 

VanderChen:
- Add pulsar client as a publisher
- Fix gRPC client to allow it coexist with the new pulsar client
- Add logic to support GS programming based on port-level configuration

### DIFF
--- a/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
+++ b/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
@@ -56,14 +56,6 @@ public class DataPlaneClientImpl implements DataPlaneClient {
         return null;
     }
 
-    private UnicastGoalStateByte buildUnicastGoalStateByte(UnicastGoalState unicastGoalState) {
-        UnicastGoalStateByte unicastGoalStateByte = new UnicastGoalStateByte();
-        unicastGoalStateByte.setNextTopic(unicastGoalState.getNextTopic());
-        unicastGoalStateByte.setGoalStateByte(unicastGoalState.getGoalState().toByteArray());
-
-        return unicastGoalStateByte;
-    }
-
     private List<String> getGroupTopics(List<String> hostIps) throws Exception {
         List<String> groupTopics = new ArrayList<>();
 
@@ -101,14 +93,6 @@ public class DataPlaneClientImpl implements DataPlaneClient {
         return multicastTopics;
     }
 
-    private MulticastGoalStateByte buildMulticastGoalStateByte(MulticastGoalState multicastGoalState) {
-        MulticastGoalStateByte multicastGoalStateByte = new MulticastGoalStateByte();
-        multicastGoalStateByte.setNextTopics(multicastGoalState.getNextTopics());
-        multicastGoalStateByte.setGoalStateByte(multicastGoalState.getGoalState().toByteArray());
-
-        return multicastGoalStateByte;
-    }
-
     private void createGoalState(MulticastGoalState multicastGoalState) throws Exception {
         Map<String, List<String>> multicastTopics = getMulticastTopics(multicastGoalState.getHostIps());
 
@@ -124,7 +108,7 @@ public class DataPlaneClientImpl implements DataPlaneClient {
                     .enableBatching(false)
                     .create();
 
-            producer.send(buildMulticastGoalStateByte(multicastGoalState));
+            producer.send(multicastGoalState.getMulticastGoalStateByte());
 
             LOG.info("Send multicastGoalState to topic:{} success, " +
                             "groupTopics: {}, unicastGoalStates: {}",
@@ -153,7 +137,7 @@ public class DataPlaneClientImpl implements DataPlaneClient {
                     .topic(topic)
                     .enableBatching(false)
                     .create();
-            producer.send(buildUnicastGoalStateByte(unicastGoalState));
+            producer.send(unicastGoalState.getUnicastGoalStateByte());
 
             LOG.info("Send unicastGoalStates to topic:{} success, " +
                     "unicastGoalStates: {}", nextTopic, unicastGoalState);
--- a/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
+++ b/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
@@ -39,7 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-@Component
+//@Component
 public class DataPlaneClientImpl implements DataPlaneClient {
     private static final Logger LOG = LoggerFactory.getLogger(DataPlaneClientImpl.class);
 
--- a/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
+++ b/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
@@ -32,6 +32,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import org.springframework.stereotype.Service;
 import org.springframework.util.StringUtils;
 
 import java.util.ArrayList;
@@ -40,6 +41,7 @@ import java.util.List;
 import java.util.Map;
 
 //@Component
+@Service("pulsar")
 public class DataPlaneClientImpl implements DataPlaneClient {
     private static final Logger LOG = LoggerFactory.getLogger(DataPlaneClientImpl.class);
 
--- a/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
+++ b/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
@@ -18,24 +18,151 @@ package com.futurewei.alcor.dataplane.client.pulsar;
 import com.futurewei.alcor.dataplane.client.DataPlaneClient;
 import com.futurewei.alcor.dataplane.entity.MulticastGoalState;
 import com.futurewei.alcor.dataplane.entity.UnicastGoalState;
+import com.futurewei.alcor.dataplane.exception.GroupTopicNotFound;
+import com.futurewei.alcor.dataplane.exception.MulticastTopicNotFound;
 import com.futurewei.alcor.schema.Goalstate;
 import com.futurewei.alcor.schema.Goalstateprovisioner;
-import org.springframework.beans.factory.annotation.Autowired;
+import com.futurewei.alcor.web.entity.dataplane.MulticastGoalStateByte;
+import com.futurewei.alcor.web.entity.dataplane.UnicastGoalStateByte;
 
+import org.apache.pulsar.client.api.Producer;
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.impl.schema.JSONSchema;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-//@Component
+@Component
 public class DataPlaneClientImpl implements DataPlaneClient {
+    private static final Logger LOG = LoggerFactory.getLogger(DataPlaneClientImpl.class);
+
+    @Autowired
+    private PulsarClient pulsarClient;
+
+    @Autowired
+    private TopicManager topicManager;
 
     @Override
     public Map<String, List<Goalstateprovisioner.GoalStateOperationReply.GoalStateOperationStatus>> createGoalStates(Goalstate.GoalState goalState, String hostIp) throws Exception {
         return null;
     }
 
+    private UnicastGoalStateByte buildUnicastGoalStateByte(UnicastGoalState unicastGoalState) {
+        UnicastGoalStateByte unicastGoalStateByte = new UnicastGoalStateByte();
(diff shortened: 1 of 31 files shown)


services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
```diff
@@ -18,24 +18,151 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 //@Component
```
xieus, line 42: I remember we had an old issue in the new DPM when we enable gRPC client and pulsar client, DPM can't launch. Not sure if it has been fixed. @chenpiaoping @VanderChen

services/data_plane_manager/src/main/resources/application.properties
```diff
@@ -13,3 +13,10 @@ grpc.threads-pool-name = grpc-thread-pool
 #logging.level.root=INFO
 
 
+mq.type=pulsar
+#####Pulsar configuration#####
+pulsar.url=pulsar://127.0.0.1:6650
+pulsar.unicast.topic=unicast-topic1
+host.ip.to.group.topic.map=group-topic1:192.168.131.131,10.10.10.11 group-topic2:192.168.131.131,11.11.11.12
```
xieus, line 20: This is an OK short-term solution. Let us plan a long-term solution. One possibility to store the mapping in node metadata manager, as this is a host-level resource.

@chenpiaoping @VanderChen what do you think of?

services/data_plane_manager/src/main/resources/application.properties
```diff
@@ -13,3 +13,10 @@ grpc.threads-pool-name = grpc-thread-pool
 #logging.level.root=INFO
 
 
+mq.type=pulsar
```
xieus, line 16: Like this config 👍

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/TopicManager.java
```diff
@@ -15,9 +15,72 @@
...
+
+@Configuration
+@ConditionalOnProperty(prefix = "mq", name = "type", havingValue = "pulsar")
 public class TopicManager {
+    private static final Logger LOG = LoggerFactory.getLogger(TopicManager.class);
+
+    @Autowired
+    private PulsarConfiguration configuration;
+
+    private Map<String, String> hostIpToGroupTopic;
```
xieus, line 35: Let us plan to build these mapping from node metadata manager and store it in the local cache under development.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/TopicManager.java
```diff
@@ -15,9 +15,72 @@
...
+
+    private Set<String> multicastTopics;
+
+    @PostConstruct
+    public void init() throws Exception {
+        try {
+            this.hostIpToGroupTopic = parseTopicConfig(configuration.getHostIpToGroupTopicMap());
+            this.groupTopicToMulticastTopic = parseTopicConfig(configuration.getGroupTopicToMulticastTopicMap());
+        }catch (Exception e) {
+            throw new Exception("Parse topic config error: " + e);
```
xieus, line 49: We can define an explicit exception for this parsing error. Please find an example in https://github.com/futurewei-cloud/alcor/tree/master/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/exception 

The idea is that we know every possible exception our own program may throw, and define it before hand. In each of the predefined exception, we can define its error code and message, and come to the debugging time, we can quickly locate where the exception is thrown and why based on the given error code and message.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/UnicastFunction.java
```diff
@@ -0,0 +1,47 @@
...
+package com.futurewei.alcor.dataplane.client.pulsar;
+
+import com.futurewei.alcor.web.entity.dataplane.UnicastGoalStateByte;
+import org.apache.pulsar.client.api.Schema;
+import org.apache.pulsar.functions.api.Context;
+import org.apache.pulsar.functions.api.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class UnicastFunction implements Function<UnicastGoalStateByte, UnicastGoalStateByte> {
```
xieus, line 25: Like this function implementation, which is neat.

xieus: btw, where is this UnicastFunction used?

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/MulticastFunction.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class MulticastFunction implements Function<MulticastGoalStateByte, MulticastGoalStateByte> {
+    private static final Logger LOG = LoggerFactory.getLogger(MulticastFunction.class);
+    public static final ThreadPoolExecutor executor = new ThreadPoolExecutor(
+            20,
```
xieus, line 35: we should use configuration here. 

Check the thread pool configurations in application.properties

```
grpc.min-threads = 100
grpc.max-threads = 200
```

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
```diff
@@ -18,24 +18,151 @@
...
+
+            String topic = nextTopic;
+            String unicastTopic = topicManager.getUnicastTopic();
+            if (!StringUtils.isEmpty(unicastTopic)) {
+                unicastGoalState.setNextTopic(nextTopic);
+                topic = unicastTopic;
+            }
+
+            Producer<UnicastGoalStateByte> producer = pulsarClient
+                    .newProducer(JSONSchema.of(UnicastGoalStateByte.class))
```
xieus, line 150: In addition to JSONSchema, what does other schema Pulsar support?

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
```diff
@@ -18,24 +18,151 @@
...
+                    .topic(topic)
+                    .enableBatching(false)
+                    .create();
+            producer.send(buildUnicastGoalStateByte(unicastGoalState));
+
+            LOG.info("Send unicastGoalStates to topic:{} success, " +
+                    "unicastGoalStates: {}", nextTopic, unicastGoalState);
+        }
+
+        List<Goalstateprovisioner.GoalStateOperationReply.GoalStateOperationStatus> tempList = new ArrayList<>();
```
xieus, line 160: I know we don't have a response from MQ for now. @chenpiaoping Let us plan to design a feedback channel here: it would be ideal if we can reuse the same topic with different key.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
```diff
@@ -18,24 +18,151 @@
...
         return null;
     }
 
+    private UnicastGoalStateByte buildUnicastGoalStateByte(UnicastGoalState unicastGoalState) {
+        UnicastGoalStateByte unicastGoalStateByte = new UnicastGoalStateByte();
+        unicastGoalStateByte.setNextTopic(unicastGoalState.getNextTopic());
+        unicastGoalStateByte.setGoalStateByte(unicastGoalState.getGoalState().toByteArray());
+
+        return unicastGoalStateByte;
+    }
```
xieus, line 63: This basically converts UnicastGoalState to UnicastGoalStateByte. Can we define it inside the class of UnicastGoalState?

xieus (changes requested):
Thank you. This is a good start. Please resolve the comments one by one, and work with @chenpiaoping if you have any question.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
VanderChen: I don't know whether pulsar and gRPC should be enabled at the same time. If not, we plan to use a configuration file to decide which client.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/TopicManager.java
VanderChen: I will add a predefined exception to solve this problem.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/UnicastFunction.java
VanderChen: Both UnicastFunction and MulticastFuncation will be bundled as jar packages. These will be deployed to the Pulsar cluster as function workers by the Pulsar CLI tool.

VanderChen: I'm trying to use Java interface to deploy function workers. This will enable automatic deployment when the dpm service starts.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/MulticastFunction.java
VanderChen: I'll fix this in the next commit.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
xieus: Configuration is one option which means that gRRC and Pulsar is exclusive to each other. 

@VanderChen, can you check how to allow two client coexist? Thank you.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/UnicastFunction.java
xieus: Cool looking forward to it 👍

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/pulsar/DataPlaneClientImpl.java
chenpiaoping: In the current implementation, there can only be one dataClient instance. Do we need two dataClient instances in one dpm at the same time? if so, which one should be used?

xieus: The short answer is yes, we would love to need two clients coexist in one DPM instance. The gPRC client is for our fast path, and the pulsar or other MQ clients is for scaling path. 

On the service layer, we could develop an algorithm to decide path to go. A simplest algorithm is purely based on the is_fast field in PortEntity but we could generate more sophisticated algorithm that consider the size of VPC (for the port to be updated), scenarios ( port/sg/route/nacl update), and other parameters.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/ovs/DataPlaneServiceImpl.java
```diff
@@ -557,12 +557,11 @@ private UnicastGoalState buildUnicastGoalState(NetworkConfiguration networkConfi
         multicastGoalState.setGoalState(multicastGoalState.getGoalStateBuilder().build());
         multicastGoalState.setGoalStateBuilder(null);
 
-        // TODO: Find a field to decide client
-        if (true){
-            return pulsarDataPlaneClient.createGoalStates(unicastGoalStates, multicastGoalState);
+        if (networkConfig.getPortEntities().get(0).isFastPath()){
```
VanderChen, line 560: I have a question about this. Will there be different isFastPath values in multiple portEntity? How to deal with this situation.

xieus: IsFastPath is a port-level config. Although we allow different values for multiple port entities for the sake of flexibility, we don't have such a scenario for now. In terms of implementation, we could make a port-level decision.

VanderChen: In this case, for the same NetworkConfiguration, we need to divide it into two groups to generate Goalstate based on whether it is fastpath. Am I following this logic to modify

chenpiaoping (approved):

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/ovs/DataPlaneServiceImpl.java
xieus: If a NetworkConfiguration includes multiple port entities, DPM will process each entity independently and likely send the corresponding GS to its target host. 

At the moment of sending the GS, we need to use the isFastPath field to determine go with the fast path or MQ scaling path.

services/data_plane_manager/src/main/resources/application.properties
xieus: Create a tracking issue https://github.com/futurewei-cloud/alcor/issues/490

@VanderChen If you can take this item, that would be great. Let us sync on Slack for more details.

xieus (approved):
LGTM. @VanderChen.

I left a few comments and create two tracking issue #489 and #490 for two pending requests. Please take a look.

PR #501, opened 2020-12-03
[Hotfix] Fix Issue that gRPC Client and Pulsar Client Can't Coexist

chenpiaoping:
By design, Data-plane Manager should allow multiple data plane clients coexist so that we could enable gRPC fast path and MQ scaling path simultaneously. This PR fixes the issue that DPM can't start (commit a893709) when two types of clients were enabled at the same time.


xieus (approved):
LGTM.

VanderChen (approved):

PR #502, opened 2020-12-03
[Enhancement] Verify Project Id for DB query

Gzure:
Verify project id in DB query when no token is passed to Alcor microservices.


xieus (approved):
LGTM

Issue #455, opened 2020-10-29
[VPC] The type qos_policy_id in the VpcEntity is incorrect.


cj-chung:
In the **VpcEntity**, `qos_policy_id` is declared as:
```
@JsonPropertiy("qos_policy_id")
private List qosPolicyId;
```

It should be changed to:

```
@JsonPropertiy("qos_policy_id")
private String qosPolicyId;
```

Issue #274, opened 2020-07-02
[Route Manager] Add transaction support to Route Repositories


kevin-zhonghao:
To support the following route repo:

- RouterRepository
- RouteTableRepository
- RouteEntryRepository
- RouteExtraAttributeRepository

xieus:
This is a very practice to create an issue to track pending items. I will assign this item to you @kevin-zhonghao

PR #514, opened 2020-12-10
[Bug Fix] Fix Issue #455 and Issue #274

kevin-zhonghao:
[Route Manager] Add transaction support to Route Repositories
https://github.com/futurewei-cloud/alcor/issues/274

[VPC] The type qos_policy_id in the VpcEntity is incorrect
https://github.com/futurewei-cloud/alcor/issues/455


lib/src/main/java/com/futurewei/alcor/common/db/repo/ICacheRepository.java
```diff
@@ -28,9 +28,9 @@
 
     Map<String, T> findAllItems(Map<String, Object[]> queryParams) throws CacheException;
 
-    void addItem(T newItem) throws CacheException;
+    void addItem(T newItem) throws CacheException, Exception;
```
xieus, line 31: @kevin-zhonghao It is generally not a good idea to change the interface because of a specific implementation.  Could you list which scenario would throw an exception rather than a cache exception?

xieus (approved):
LGTM

Issue #524, opened 2020-12-23
[PM - DPM] Fail to create a port


kevin-zhonghao:
**Description:**
Fail to create a port by using the same payload as before.

PM throw Exception:
```
[{"timestamp":"2020-12-23T00:21:28.207+0000","status":500,"error":"Internal Server Error","message":"No message available","path":"/network-configuration"}]] with root cause

DPM throw Exception:
java.lang.NullPointerException: null
	at com.futurewei.alcor.dataplane.service.impl.DpmServiceImpl.processNetworkConfiguration(DpmServiceImpl.java:350) ~[classes/:na]
	at com.futurewei.alcor.dataplane.service.impl.DpmServiceImpl.updateNetworkConfiguration(DpmServiceImpl.java:377) ~[classes/:na]
	at com.futurewei.alcor.dataplane.controller.DpmController.updateNetworkConfiguration(DpmController.java:48) ~[classes/:na]
```
---
**Payload:**
```
{
  "port": {
    "admin_state_up": true,
    "allowed_address_pairs": [
      {
        "ip_address": "11.11.11.1",
        "mac_address": "00-AA-BB-15-EB-3F"
      }
    ],
    "binding:host_id": "node2",
    "binding:vif_details": {},
    "create_at": "string",
    "description": "string",
    "device_id": "8182a4d4-ffff-4ece-b3f0-8d36e3d00070",
    "device_owner": "compute:nova",
    "dns_assignment": {},
    "dns_domain": "string",
    "dns_name": "string",
    "extra_dhcp_opts": [
      {
        "ip_version": "string",
        "opt_name": "string",
        "opt_value": "string"
      }
    ],
    "fast_path": true,
    "fixed_ips": [
      {
        "ip_address": "10.0.0.25",
        "subnet_id": "9192a4d4-ffff-4ece-b3f0-8d36e3d88000"
      }
    ],
    "id": "9192a4d4-ffff-4ece-b3f0-8d36e3d00112",
    "mac_learning_enabled": true,
    "name": "port108",
    "network_id": "9192a4d4-ffff-4ece-b3f0-8d36e3d88038",
    "network_ns": "string",
    "port_security_enabled": true,
    "project_id": "3dda2801-d675-4688-a63f-dcda8d327f50",
    "qos_network_policy_id": "string",
    "qos_policy_id": "string",
    "revision_number": 0,
    "security_groups": [
      "3dda2801-d675-4688-a63f-dcda8d111111"
    ],
[...]

PR #525, opened 2020-12-25
[]Fix the problem of failed to create a port

chenpiaoping:
Fix issue: https://github.com/futurewei-cloud/alcor/issues/524


xieus (approved):
LGTM.

PR #534, opened 2021-01-07
[Fundamental] Fix Java Version in Common Lib

xieus:
Revert the java version downgrade and get back to 11 in common lib.


cj-chung (approved):
LGTM

Issue #527, opened 2020-12-29
[Micro-service] Integration with PM and DPM


kevin-zhonghao:
There are some code issues in integration with PM and DPM.

1. DataPlaneManagerRestClient:

@DurationStatistics
    public void deleteNetworkConfig(NetworkConfiguration message) throws Exception {
        HttpEntity<NetworkConfiguration> request = new HttpEntity<>(message);
        restTemplate.put(dataPlaneManagerUrl, request);
    }

**Should we use restTemplate.delete() instead of restTemplate.put()?**

---

2. DpmController:

@DeleteMapping({"/network-configuration", "v4/network-configuration"})
    @DurationStatistics
    public InternalDPMResultList deleteNetworkConfiguration(@RequestBody NetworkConfiguration networkConfiguration) throws Exception {
        checkNetworkConfiguration(networkConfiguration);
        return dpmService.deleteNetworkConfiguration(networkConfiguration);
    }

**We may not use @RequestBody in DELETE Rest API**

---

3. when integration of PM and DPM failed (we are creating a normal port), it needed to be rolled back, but the port just created still existed in db. 
**We may delete port when integration failed**
Then when I tried to delete the port, it show error - **"Non existing MAC address."**

PR #537, opened 2021-01-09
[Integration] Fix PM/DPM Port Issue tracked by #527

chenpiaoping:
Since the delete method of restTemplate does not support carrying body, the put method is used to send delete message.


chenpiaoping:
@xieus @kevin-zhonghao please review this PR if you have a chance.

xieus (approved):
LGTM

PR #538, opened 2021-01-14
Fix basicTest.sh. /project/ was missing in the subnet query URL.

pkommoju:
basicTest.sh was failing to GET subnet information because the URL did not specify the "project" endpoint/path.


xieus (approved):
LGTM.

PR #539, opened 2021-01-14
[Schema][DPM] Update protobuf contract and opType

cj-chung:
This PR includes:
1. Update vpc.proto (remove subnetid)
2. Update the order of OperationType in common.proto
3. Unbind VpcConfiguration and AuxGateway in vpc.proto
4. Remove auxiliarygateway.proto
5. Create a new gateway.proto
6. remove all related codes in DPM about the contract change.
7. fix the opType for vpc and subnet when a port is created, updated, and deleted in DPM.


schema/proto3/common.proto
```diff
@@ -29,17 +29,15 @@ enum ResourceType {
...
 enum OperationType {
-    CREATE = 0;
-    UPDATE = 1;
-    GET = 2;
-    DELETE = 3;
-    INFO = 4;
-    NEIGHBOR_CREATE_UPDATE = 5; // To be removed after using NeighborState
-    NEIGHBOR_GET = 6;           // To be removed after using NeighborState
-    NEIGHBOR_DELETE = 7;        // To be removed after using NeighborState
+    INFO = 0;
```
er1cthe0ne, line 36: looks good. I am glad we can remove the legacy NEIGHBOR_* OperationTypes now.

schema/proto3/gateway.proto
```diff
@@ -19,15 +19,18 @@ syntax = "proto3";
...
 
 option java_package = "com.futurewei.alcor.schema";
-option java_outer_classname = "AuxiliaryGateway";
+option java_outer_classname = "Gateway";
 
-enum AuxGatewayType {
+enum GatewayType {
     NONE = 0;
     ZETA = 1;
+    TGW = 2;
```
er1cthe0ne, line 27: it will be great if we can put comments on what TGW/IGW/NGW stands for.

schema/proto3/gateway.proto
```diff
@@ -19,15 +19,18 @@ syntax = "proto3";
 package alcor.schema;
 
 option java_package = "com.futurewei.alcor.schema";
-option java_outer_classname = "AuxiliaryGateway";
+option java_outer_classname = "Gateway";
 
-enum AuxGatewayType {
+enum GatewayType {
```
er1cthe0ne, line 24: I was calling it "AuxiliaryGateway" to no confused with general networking "Gateway" for cross subnet communication. I guess I will accept we calling it as "Gateway" since most public cloud provider are calling it this way. :)

schema/proto3/gateway.proto
```diff
@@ -19,15 +19,18 @@ syntax = "proto3";
...
+    IGW = 3;
+    NGW = 4;
 }
 
-message AuxGateway {
-    AuxGatewayType aux_gateway_type = 1;
+message GatewayConfiguration {
+    GatewayType gateway_type = 1;
 
     // if aux_gateway_type == NONE, then id = empty string
```
er1cthe0ne, line 35: comment update: aux_gateway_type to gateway_type

schema/proto3/gateway.proto
```diff
@@ -19,15 +19,18 @@ syntax = "proto3";
...
     ZETA = 1;
+    TGW = 2;
+    IGW = 3;
+    NGW = 4;
 }
 
-message AuxGateway {
-    AuxGatewayType aux_gateway_type = 1;
+message GatewayConfiguration {
+    GatewayType gateway_type = 1;
```
er1cthe0ne, line 33: since we moved Gateway as a top_level resource, we should add the following in the begining similar to other top level resources:

    uint32 revision_number = 1;

    string request_id = 2;
    string id = 3;        *** note that I move this up for consistancy
    UpdateType update_type = 4; // DELTA (default) or FULL

schema/proto3/vpc.proto
```diff
@@ -34,14 +34,12 @@ message VpcConfiguration {
...
+/*
     message SubnetId {
         string id = 1;
     }
-
-    repeated SubnetId subnet_ids = 9;
-
-    AuxGateway auxiliary_gateway = 10;
+*/
+    repeated string auxiliary_gateway_ids = 9;
```
er1cthe0ne, line 42: auxiliary_gateway_ids - > gateway_ids

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/SubnetService.java
```diff
@@ -85,7 +85,7 @@ public void buildSubnetStates(NetworkConfiguration networkConfig, UnicastGoalSta
             }
 
             Subnet.SubnetState.Builder subnetStateBuilder = Subnet.SubnetState.newBuilder();
-            subnetStateBuilder.setOperationType(networkConfig.getOpType());
+            subnetStateBuilder.setOperationType(Common.OperationType.INFO); // networkConfig.getOpType()
```
er1cthe0ne, line 88: please remove the commented out part.

actually, you don't even need to set it to INFO because it is the default based on your schema update. I suggest removing this to save this one line of execution :) and replace with a comment it is set as INFO by default.

xieus: As discussed, let us create a standalone class to assign the OperationType in various scenarios, and pass the reference of the object to VPC/Subnet/.. services classes to consume the value. 

This way, we have a centralized place for field assignment and make it easier to change in the future.

schema/proto3/gateway.proto
cj-chung: comments added

schema/proto3/gateway.proto
cj-chung: more fields in the gateway contract are added.

schema/proto3/gateway.proto
cj-chung: fixed

schema/proto3/vpc.proto
cj-chung: fixed

er1cthe0ne (approved):
LGTM

xieus (approved):

PR #542, opened 2021-01-19
service design

issacyxw:


xieus (approved):
We've discussed the design in the open-source meeting. Merged to catch up.

PR #554, opened 2021-03-05
[Schema] Add Id field in DHCP state

zzxgzgz:
This PR addes a field, `string id = 3;`, to dhcp.proto. 

It has been tested and the unit tests passed. Please refer to the following screenshot.

image


xieus (approved):
Looks good.

PR #556, opened 2021-03-10
Add NCM_URI and NCM_ID to NodeInfo

pkommoju:
Orignally only ncm_uri was to be added but in a discussion with Liguang,
it became apparent that a "ncm_id" will be required and since adding it
elsewhere would require a join, it is added to NodeInfo.

### DIFF
--- a/web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
+++ b/web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
@@ -71,19 +71,19 @@ public class NodeInfo implements Serializable {
         this(nodeInfo.id, nodeInfo.name, nodeInfo.localIp, nodeInfo.macAddress, nodeInfo.veth, nodeInfo.gRPCServerPort, nodeInfo.hostDvrMac, nodeInfo.unicastTopic, nodeInfo.multicastTopic, nodeInfo.groupTopic, nodeInfo.ncm_uri, nodeInfo.ncm_id);
     }
 
-    public NodeInfo(String id, String name, String localIp, String macAddress, String veth, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic, String ncm_uri, String ncm_id) {
-        this(id, name, localIp, macAddress, unicastTopic, multicastTopic, groupTopic, ncm_uri, ncm_id);
+    public NodeInfo(String id, String name, String localIp, String macAddress, String veth, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic) {
+        this(id, name, localIp, macAddress, unicastTopic, multicastTopic, groupTopic);
         this.veth = veth;
         this.gRPCServerPort = gRPCServerPort;
     }
 
-    public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic, String ncm_uri, String ncm_id) {
-        this(nodeId, nodeName, ipAddress, macAddress, unicastTopic, multicastTopic, groupTopic, ncm_uri, ncm_id);
+    public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic) {
+        this(nodeId, nodeName, ipAddress, macAddress, unicastTopic, multicastTopic, groupTopic);
         this.veth = "";
         this.gRPCServerPort = gRPCServerPort;
     }
 
-    public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, String unicastTopic, String multicastTopic, String groupTopic, String ncm_uri, String ncm_id) {
+    public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, String unicastTopic, String multicastTopic, String groupTopic) {
         this.id = nodeId;
         this.name = nodeName;
         if (this.validateIp(ipAddress)) {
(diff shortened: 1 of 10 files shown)


web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
```diff
@@ -57,27 +57,33 @@
...
+    public NodeInfo(String id, String name, String localIp, String macAddress, String veth, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic, String ncm_uri, String ncm_id) {
+        this(id, name, localIp, macAddress, unicastTopic, multicastTopic, groupTopic, ncm_uri, ncm_id);
         this.veth = veth;
         this.gRPCServerPort = gRPCServerPort;
     }
 
-    public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic) {
-        this(nodeId, nodeName, ipAddress, macAddress, unicastTopic, multicastTopic, groupTopic);
+    public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic, String ncm_uri, String ncm_id) {
+        this(nodeId, nodeName, ipAddress, macAddress, unicastTopic, multicastTopic, groupTopic, ncm_uri, ncm_id);
```
cj-chung, line 81: @pkommoju I think you can create a new overloaded constructor NodeInfo(....) to include all parameters and two new parameters you added. Don't edit any existing constructors. Inside the constructor, you can use this() parent constructor to initialize the matching parameters and use another two statements to assign values for two new parameters.

cj-chung (changes requested):

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
```diff
@@ -57,27 +57,33 @@
     @JsonProperty("group_topic")
     private String groupTopic;
 
+    @JsonProperty("ncm_uri")
+    private String ncm_uri;
```
xieus, line 61: @pkommoju If url is part of NodeInfo, in the case that we NCM service url is updated, we will need to modify all impacted nodes. 

Another option is to create a new KV cache, ncm_id => ncm_info. Think about this option.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
xieus: +1

pkommoju: James,
I am looking into this.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
pkommoju: I am looking into it.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
pkommoju: After going through all calls to NodeInfo constructor, I realized that most variations of constructors are inside NodeInfo itself, which will be taken care by your suggestion, in addition to changing this constructor, which is sort of common constructor to set ncm_uri and ncm_id to null.

>>> public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, String unicastTopic, String multicastTopic, String groupTopic)

But then there are these other cases from outside NodeInfo:
NodeFileLoader, NodeControllerTest, MockRestClientAndRepository all use

NodeInfo(strId, strName, strIp, strMac, strVeth, nServerPort, strUnicastTopic, strMulticastTopic, strGroupTopic);

They need to pass ncm_uri, ncm_id. So I had to add another override for this constructor.

I have a question about style/convention here. Is it preferred to set them to "" or null?

I need to refresh my branch and build and test.

cj-chung: You just need to create a new constructor like:
```
public NodeInfo(String nodeId, String nodeName, String ipAddress, String macAddress, int gRPCServerPort, String unicastTopic, String multicastTopic, String groupTopic, String ncm_uri, String ncm_id) {
        this(nodeId, nodeName, ipAddress, macAddress, unicastTopic, multicastTopic, groupTopic);
        this.ncm_uri = ncm_uri;
        this.ncm_id = ncm_id;
}
```
You don't need to modify other constructors, I think.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
pkommoju: I think the deciding factor could be frequency of updating URI vs frequecny of having to join. Say, a given URI_NCM_one is on NCM_one, which is managing node_000 through node_999 (1000 nodes). If URI_one of NCM_one is changed to URI_NCM_one_new, then 1000 nodeinfo will have to be updated every time URI_NCM_xyz changes.
On the otherhand, DPM/NCM will have to take the node_info.ncm_id and join (query ncm_id => ncm_info) to get the NCM URI.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
pkommoju: If the basecase like constructor is not modifed, what will the values of ncm_uri and ncm_id? Won't they be garbage?

cj-chung: if there is no initialization for a data member in the constructor, that data member will be null. You can use set method to set value for data members too.

pkommoju: Well, then, C++ is still lingering in my head. I will get rid of the other constructor changes.

PR #559, opened 2021-03-10
[Schema] Support multiple same-type resources per host

xieus:
This PR provides a simple update of GoalState V2 schema to support multiple same-type resources per host.


er1cthe0ne (approved):
LGTM

PR #563, opened 2021-03-12
[Node Metadata Mgr] Add NCM Id field to NodeInfo

pkommoju:
NMM changes - part 1
* Add ncm_id to NodeInfo.
* Use setNcmId method to set the ncm_id after creating NodeInfo because
otherwise calls to overloaded constructors are resolving to incorrect
ones, or resulting in errors such as constructor is already defined, or
can't resolve constructor.

### DIFF
--- a/services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/NodeService.java
+++ b/services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/NodeService.java
@@ -26,4 +26,8 @@ public interface NodeService {
     NodeInfo updateNodeInfo(String nodeId, NodeInfo nodeInfo) throws ParameterNullOrEmptyException, InvalidDataException, Exception;
 
     String deleteNodeInfo(String nodeId) throws ParameterNullOrEmptyException, Exception;
+ 
+    // TEMP: URI will be provided by the Admin but these are for testing purpose.
+    static String makeUpNcmId(String hostIP, int localPort) { return "ncm_" + hostIP + "_" + String.valueOf(localPort); }
+    static String makeUpNcmUri(String hostIP, int localPort) { return "ncm/" + hostIP + "/" + String.valueOf(localPort); }
 }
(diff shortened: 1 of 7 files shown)


services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/NodeService.java
```diff
@@ -26,4 +26,8 @@
     NodeInfo updateNodeInfo(String nodeId, NodeInfo nodeInfo) throws ParameterNullOrEmptyException, InvalidDataException, Exception;
 
     String deleteNodeInfo(String nodeId) throws ParameterNullOrEmptyException, Exception;
+ 
+    // TEMP: URI will be provided by the Admin but these are for testing purpose.
+    static String makeUpNcmId(String hostIP, int localPort) { return "ncm_" + hostIP + "_" + String.valueOf(localPort); }
```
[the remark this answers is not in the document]
pkommoju: I realized that creating an ID is required. However, it cannot be tied to any specific node, as you noted. So, this will be makeUpNcmId will be made official but I think I will find a better name, something like create/generateNcmId.

services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/NodeService.java
```diff
@@ -26,4 +26,8 @@
     NodeInfo updateNodeInfo(String nodeId, NodeInfo nodeInfo) throws ParameterNullOrEmptyException, InvalidDataException, Exception;
```
cj-chung, line 26: @pkommoju Is that possible for a node migrating to another ncm?

xieus (approved):
LGTM with one pending item. Please take a look.

services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/NodeService.java
pkommoju: We haven't thought about it, but I think it could be supported by a delete in one NCM followed by add to the other NCM. I am not sure how this would affect DPM and NMM. If there is no strong enough reason, we might be better off to not support it, at least at this time.

PR #564, opened 2021-03-16
[Data-Plane Mgr] Fix missing subnet entity in multicast GoalState

xieus:
__Bug Description__

Each port creation generates one unicast GS and one multicast GS. In current implementation, multicast GS only includes neighbor states while lacks of corresponding subnet states. This causes ACA failure to look up neighbor.

This PR proposes a simple fix to add those missing subnet entities, and enable debug-level logging to show GS detail before sending it to ACA.


er1cthe0ne (approved):
LGTM

PR #570, opened 2021-03-24
[Network Config Mgr] Add NodeInfo APIs

pkommoju:
NodeInfo cache in NCM is working.
Bulk API is not tested.

### DIFF
--- a/scripts/ncmtest.sh
+++ b/scripts/ncmtest.sh
@@ -1,0 +1,25 @@
+#! /bin/sh
+
+# Expect to fail
+echo "DELETE Existing node"
+curl -X DELETE -H "Content-Type: application/json" http://localhost:9014/nodes/9192a4d4-ffff-4ece-b3f0-8d36e3d85002
+echo "GET from empty cache"
+curl -X GET -H "Content-Type: application/json" http://localhost:9014/nodes/9192a4d4-ffff-4ece-b3f0-8d36e3d85002
+echo "POST, new node"
+curl -X POST -H "Content-Type: application/json" --data @json/createNode.json http://localhost:9014/nodes
+echo "GET it back"
+curl -X GET -H "Content-Type: application/json" http://localhost:9014/nodes/9192a4d4-ffff-4ece-b3f0-8d36e3d85002
+echo "DELETE again node"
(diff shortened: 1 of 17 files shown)


services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/util/NetworkConfigManagerConstants.java
```diff
@@ -0,0 +1,5 @@
+package com.futurewei.alcor.netwconfigmanager.util;
+
+public class NetworkConfigManagerConstants {
```
xieus, line 3: Regarding the naming, I would recommend to make it short, for example, Constants. 

The full class name include NetworkConfigManager in the path.

scripts/ncmtest.sh
```diff
@@ -0,0 +1,25 @@
+#! /bin/sh
+
```
xieus, line 2: Could we create a new folder /scripts/test and move this script there?

xieus (approved):
LGTM with two minor comments.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/util/NetworkConfigManagerConstants.java
pkommoju: You mean, rename the class to Constants? I think that makes sense. I will address in the next commit.

scripts/ncmtest.sh
pkommoju: Makes sense. I will do that.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/util/NetworkConfigManagerConstants.java
xieus: Right.

scripts/ncmtest.sh
pkommoju: Are you waiting me to fix these in this branch/PR or can I do it after you merge into master?

xieus:
This PR has some Github merge issue. A new PR #572 is created.

Issue #103, opened 2020-02-04
Support deletion port E2E scenario


xieus:
ACA change is completed by this PR (https://github.com/futurewei-cloud/alcor-control-agent/pull/166). We will need PM to determine if it is a port delete for both scenarios below and then send down the port delete with the corresponding neighbor delete (same logic as port create and send down neighbor table). See here for context: https://github.com/futurewei-cloud/alcor-control-agent/pull/166#discussion_r527439986

**first scenario to support VM creation without specifiying a port, then delete the VM**

1. Create a VM giving it a created subnet but not port
1. Neutron will create a port for the new VM
1. Delete the above VM
1. Controller receives /delete request from clients and will issue a DELETE GS to ACA
1. Action item - needs to confirm DPM can send down the corresponding delete port to ACA
1. Action item - ACA add support for port delete operation

**second scenario to support VM created with specifiying a port, then delete the VM**

1. Create new port on the horizon UI
1. Create a VM using the new port above
1. Delete the above VM
1. Controller receives a /update request from clients with no device_id and device_owner, and controller passes the update request to ACA
1. Action item - needs to update port.proto to include device ID and device owner fields (status: DONE)
1. Action item - needs to confirm DPM can send down the corresponding update port to ACA (status: DONE, PR #571)
1. Action item - ACA add support for port update with device ID and device owner fields = null (status: DONE, ACA PR #227)

xieus:
I think we need to plan to fill out this gap. It has been a while. @er1cthe0ne

er1cthe0ne:
yes, we need to implement deletion of endpoint in ACA. I will look into it.

ACA only track ports but not VPC/Subnet creation/deletion. Therefore, it doesn't need to worry about VPC/Subnet create/delete.

xieus:
Correct, this issue was created previously for E2E scenarios, and it seems VPC/Subnet deletion has been covered.

er1cthe0ne:
> Correct, this issue was created previously for E2E scenarios, and it seems VPC/Subnet deletion has been covered.

yes, the ACA port delete/update is also tracked by below, I have raised the priority on it.

https://github.com/futurewei-cloud/alcor-control-agent/issues/115

xieus:
> > Correct, this issue was created previously for E2E scenarios, and it seems VPC/Subnet deletion has been covered.
> 
> yes, the ACA port delete/update is also tracked by below, I have raised the priority on it.
> 
[...]

PR #571, opened 2021-03-25
[Data Plane Mgr] Add device id and owner in PortService

xieus:
This PR fixes two missing fields in PortService at DPM so that we could support port creation and port deletion. 

Tracking issue: #103


cj-chung:
LGTM

cj-chung (approved):

PR #576, opened 2021-03-29
[Data-Plane Mgr] Fix Port Operation Type Mismatch Issue

xieus:
This PR proposes temporary fix for the following scenarios:
- Nova sends port update operation with nonempty device_id and device_owner (of value compute:nova). This request should be treated as CREATE from Alcor perspective.
- Nova sends port update operation with empty device_id and device_owner. This request should be treated as DELETE (tracked by Issue #103).

The ultimate fix should be in Port Manager as the resource owner. We proposes this fix as a short-term solution to verify E2E success.


cj-chung (approved):
LGTM

PR #587, opened 2021-04-08
[Port Mgr] Fix the status return from port update for Nova

cj-chung:
Nova doesn't like the status return from port update `"status"="success"`. 
Nova/Openstack only identify the following status from port operation:
`ACTIVE`
`BUILD`
`DOWN`
`ERROR`


xieus (approved):
LGTM.

PR #593, opened 2021-04-10
[Fix Bug] Add device id and device owner in DPM UT

VanderChen:


services/data_plane_manager/src/main/resources/application.properties
```diff
@@ -19,7 +19,7 @@ protobuf.goal-state-message.version = 102
 #logging.level.root=INFO
 
 #####Ignite configuration######
-ignite.host=localhost
+ignite.host=172.16.50.5
```
xieus, line 22: Please revert this change. You can keep this in your local branch but usually local environment setup parameters are not recommended to check in to common repository.

services/data_plane_manager/src/main/resources/application.properties
```diff
@@ -30,7 +30,7 @@ ignite.thin.client.enable=true
 
 mq.type=pulsar
 #####Pulsar configuration#####
-pulsar.url=pulsar://127.0.0.1:6650
+pulsar.url=pulsar://172.16.50.5:6650
```
xieus, line 33: Same comment as above.

xieus (approved):
@VanderChen LGTM with minor comments. Please revert the change and let us know so that I will merge the change quickly.

PR #596, opened 2021-04-14
Node Manager Interface for NCM Registration.

pkommoju:
* NMM has the interface to register, unregister, query NCM metadata.
* NMM augments NodeInfo with NCM URI based on the NCM ID in it during node registration.
* NCM Metadata in NMM mo longer carries list of node_ids.
* NMM swagger UI allows querying nodes by NCM ID.

Known problems:
* Node creation is not returning error the ncm_id has not been registered.
* NMM query by ncm_id is not working.
* NMM is not using NCM URI to talk to NCM.
* No Unit tests yet, only swagger UI has been used to validate the code.


pkommoju:
This is a rather large change set.
If any existing test/code has cases where a node is created with a ncmid set but that ncm has not been registered, it may fail.

xieus (approved):
LGTM

PR #598, opened 2021-04-15
[Node Mgr] Node Manager Interface for NCM Registration

pkommoju:
Node Manager Interface for NCM Registration

__IMPORTANT NOTE__
* This is yet another breaking change. This commit introduces named
versions of NodeInfo cache in NMM, DPM and NCM. Previously, the three
were physically the same cache.

If any code/test assumes the existence of certain node in these caches
before this change, they will not find those entries since this change.
creates new caches.

* This commit also introduces an implicit assumption that bulk adding of
nodes can only be done to one NCM at a time. Even if each nodeinfo has a
distinct NCM id, NMM will pick the NCM Id from the first nodeinfo and
send the whole batch to that NCM (issue #601 is tracking this)

* Node creation via file upload was meant to be an admin task and was
not designed to dispatch those nodes to DPM and NCM. This commit
does not change this old behavior (issue #600 is tracking this)

Regular change notes:

* Node creation will continue to work if nodeinfo has no ncmid to be backward
compatible. Delete, update file upload, bulk add all are 100% backward
compatible if ndmid is missing.

* If ncmid is present, it must resolve to a registered NCM or an error will
be raised.

* Update integration test script (scripts/test/nmm-ncm-test.sh) with NCM
* id/uri. UPDATE is also validated. Added node creation with no NCMID.

* Fix update/delete getting called through wrong methods in DPM/NCM rest
client code. Using rest template exchange method with correct HTTP
headers is required in those sites.


xieus:
> IMPORTANT NOTE:
> This is a breaking change in the following way:
> create/update node will fail if NodeInfo doesn't contain a valid
> NCM uri.
> This means, 1) An instance of NCM must be running, 2) The NCM
> should be registered with NMM first. NCM Registration should
> contain the valid http end-point (like, http://localhost:9014)
> and an ID. NodeInfo should contain this NCM ID.

Good to make a breaking change explicit. Do we support  node registration with an empty or null NCM id?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
```diff
@@ -85,7 +85,14 @@ public static void close() {
 
     public static Ignite getIgnite(){
         // if no need create a real ignite server, we return a mock Ignite client
+        // NOT_JAVA_11_COMPATIBLE: doesn't work in IntelliJ IDE
         return Objects.requireNonNullElseGet(igniteServer, IgniteNodeClientMock::new);
```
cj-chung, line 89: @pkommoju line 89 should work if you set Java version to 11 in Intellij

lib/src/main/java/com/futurewei/alcor/common/rbac/StrictRbacManager.java
```diff
@@ -122,10 +122,14 @@ public void checkCreate(String resourceName, TokenEntity tokenEntity, List<Strin
     private List<String> excludeFields(String actionName, TokenEntity tokenEntity, OwnerChecker ownerChecker) {
         List<String> excludeFields = new ArrayList<>();
         Optional<ActionRbacRule> rbacRuleOptional = serviceRbacRule.getActionRbacRule(actionName);
+        // NOT_JAVA_11_COMPATIBLE: doesn't work in IntelliJ IDE
         if (rbacRuleOptional.isEmpty()) {
```
cj-chung, line 126: @pkommoju same as here, you need to set Java to 11.

services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/implement/NodeServiceImpl.java
```diff
@@ -100,30 +100,21 @@ private void handleCreateNodeBulkRequest(List<NodeInfo> nodeInfos) {
         }
     }
 
-    private boolean augmentNodeInfosWithNcmUri(List<NodeInfo> nodeInfos) {
+    private void augmentNodeInfosWithNcmUri(List<NodeInfo> nodeInfos) throws Exception {
         for (int i = 0; i < nodeInfos.size(); ++i) {
             NodeInfo ni = nodeInfos.get(i);
-            if (addNcmUriToNodeInfo(ni) == false)
-                return false;
+            addNcmUriToNodeInfo(ni);
```
cj-chung, line 106: @pkommoju why you need to loop through every node here?

cj-chung: @pkommoju I see, this is for node file upload.

web/src/main/java/com/futurewei/alcor/web/restclient/NetworkConfigManagerRestClient.java
```diff
@@ -45,18 +45,33 @@
...
     @DurationStatistics
     public void deleteNodeInfo(String nodeId) throws Exception {
+        /*
+         * PROBLEM: To dispatch node deletion to the correct NCM, we need it's
+         * URI but we can't get it here.
+         * FIX: Change the API to pass in NodeInfo.
+         * String ncmUrl = message.getNodeInfo().getNcmUri();
+         *
+            if (ncmUrl == null)
+                throw new Exception("NetworkConfigManagerClient: Required ncm_uri is NULL");
```
cj-chung, line 73: @pkommoju why not search ncm_id from nodeIno if you have nodeId here? Then you can remove the node from the ncm.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/MockIgniteServer.java
pkommoju: It gives me runtime error only if I try to debug in the IDE.

lib/src/main/java/com/futurewei/alcor/common/rbac/StrictRbacManager.java
pkommoju: same problem.

services/node_manager/src/main/java/com/futurewei/alcor/nodemanager/service/implement/NodeServiceImpl.java
pkommoju: Every nodeinfo needs to have it's ncmid resolved to correct ncmuri.

web/src/main/java/com/futurewei/alcor/web/restclient/NetworkConfigManagerRestClient.java
pkommoju: The method gets called with nodeId. What is in the comment is just a note to self.

web/src/main/java/com/futurewei/alcor/web/restclient/NetworkConfigManagerRestClient.java
```diff
@@ -45,25 +45,49 @@
 
     @DurationStatistics
     public void createNodeInfo(NodeInfoJson message) throws Exception {
+        String ncmUri = message.getNodeInfo().getNcmUri();
+        if (ncmUri == null)
+            ncmUri = ncmManagerUrl;
```
xieus, line 50: Maybe we want to rename "ncmManagerUrl" to "defaultNcmUrl"?

xieus (approved):
LGTM.

xieus:
@pkommoju Please create a tracking issue for every outstanding item.

web/src/main/java/com/futurewei/alcor/web/restclient/NetworkConfigManagerRestClient.java
pkommoju: Do you want me to make the change and push into this PR, or will you do it in your merge?

xieus: Let us make it in this PR.

xieus (commented):
New latest change looks good!

PR #606, opened 2021-05-03
[Test Script] Update busybox ping test to run from Jenkins

kiran1048:
Added features to restart ACA on remote nodes before start of test.
All tests being on AWS machines.

### DIFF
--- a/scripts/busybox-ping-test/container_ops.py
+++ b/scripts/busybox-ping-test/container_ops.py
@@ -22,14 +22,25 @@ from termcolor import colored
 
 def busybox_container_cleanup(ip, con):
     print("Cleaning up busybox container", con)
-    command = "ovs-docker del-ports br-int ".format(con)
+    command = "sudo ovs-docker del-ports br-int ".format(con)
     output = run_command_on_host(ip, command)
-    command = "docker container stop ".format(con)
+    print("CCC",command,output)
+    command = "sudo docker container stop {}".format(con)
     output = run_command_on_host(ip, command)
-    command = "docker container rm {} -f ".format(con)
+    print("CCC",command,output)
+    command = "sudo docker container rm {} -f ".format(con)
     output = run_command_on_host(ip, command)
-
-
+    print("CCC",command,output)    
+    command = "sudo ovs-vsctl del-br br-tun"
+    output = run_command_on_host(ip,command)
+    print("CCC",command,output)    
+    command = "sudo ovs-vsctl del-br br-int"
+    output = run_command_on_host(ip,command)
+    print("CCC",command,output)
+    command = "sudo /usr/local/share/openvswitch/scripts/ovs-ctl start"
+    output = run_command_on_host(ip,command)
+    print("CCC",command,output)
+    
 def busybox_container_deploy(target_ips, ip_mac_db, container_names):
     print("Deploying busybox container")
     index = 0;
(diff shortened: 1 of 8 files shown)


scripts/busybox-ping-test/container_ops.py
```diff
@@ -1,62 +1,78 @@
...
+    print("Cleanup task: ", command, "\n", output, "\n")
+    command = "sudo ovs-vsctl del-br br-tun"
+    output = run_command_on_host(aca_node_ip, command)
+    print("Cleanup task: ", command, "\n", output, "\n")
+    command = "sudo ovs-vsctl del-br br-int"
+    output = run_command_on_host(aca_node_ip, command)
+    print("Cleanup task: ", command, "\n", output, "\n")
+    command = "sudo /usr/local/share/openvswitch/scripts/ovs-ctl start"
+    output = run_command_on_host(aca_node_ip, command)
+    print("Cleanup task: ", command, "\n", output, "\n")
```
pkommoju, line 42: Is there a validation of output from run_command_on_host(), say, if it suceeded or failed and bail out if it had failed? This will prevent us going over more log liens than necessary.

scripts/busybox-ping-test/ping_test.py
```diff
@@ -66,55 +89,79 @@ def start_containers(serv):
...
+    aca_nodes_ip_mac = get_macaddr_alcor_agents(aca)
+    print("ACA nodes IP MAC pair::", aca_nodes_ip_mac)
+
+    if len(aca_nodes_ip_mac) != len(aca):
+      print("ERROR: Alcor Control Agent not running on some of the nodes")
+      print("ERROR: Quitting test\n")
+      sys.exit(1)
+
+    print("Wait for 60 seconds until all services are started...")
+    time.sleep(60)
```
pkommoju, line 164: May be wait in a loop for a set number of times, each time testing if all services are up and running by looking at netstat -ant and count the number of alcor ports that should be in LISTEN mode could be better.

scripts/busybox-ping-test/container_ops.py
kiran1048: We don't check here. But in the next step we do check if ACA is up and running on the target nodes. If for any reason we don't find ACA running then test stops by giving a message that ACA is not running.

scripts/busybox-ping-test/ping_test.py
kiran1048: We can address this in an upcoming commit.

scripts/busybox-ping-test/helper_functions.py
```diff
@@ -60,18 +105,36 @@ def get_mac_id(HOST):
     mac_addr = addr_string[addr_string.index('ether') + 1]
     return mac_addr
 
+def get_username():
+   #return pwd.getpwuid( os.getuid() )[ 0 ]
+   return 'ubuntu'
+
```
pkommoju, line 111: I forgot to ask about this change. What does the user ubuntu signify here, it is not the logged in user, "jenkins" user. Waht is this return value used for?

pkommoju (approved):
A clarification about get_username(), ubuntu, and SSH would be very helpful.

scripts/busybox-ping-test/helper_functions.py
kiran1048: You are right, the script uses this user 'ubuntu' to log on to the target machine to carry on its operations. The module pwd.getpwuid() return 'jenkins' when run from jenkins. So we would then need the jenkin's public key to be copied over to our targets to work. Since ubuntu is a default user that we use on all three hosts in questions, so it makes it easier to run the tests using that account.

kiran1048:
> A clarification about get_username(), ubuntu, and SSH would be very helpful.

yes, in the test setup readme, we will keep the steps needed to allow the three hosts to seamlessly ssh to each other in order to run the tests.

scripts/busybox-ping-test/container_ops.py
```diff
@@ -1,62 +1,78 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 from helper_functions import *
 from termcolor import colored
 
-def busybox_container_cleanup(ip, con):
+def busybox_container_cleanup(aca_node_ip, con):
```
xieus, line 23: Like the change as making the parameter explicit. 

Btw, con = container? Can we also make a similar change here?

xieus (approved):
LGTM. only one minor comment. Thank you @kiran1048.

PR #607, opened 2021-05-03
[Network Config Mgr] Support On Demand Service

xieus:
This PR adds the following features:
* Add new on-demand service in the new Network Configuration Manager
* Based on on-demand request from the host, implement on-demand algorithm to retrieve from NCM cache and send down GoalState to ACA, as well as to respond to ACA's request


services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/cache/VpcResourceCache.java
```diff
@@ -28,7 +27,7 @@
 @ComponentScan(value = "com.futurewei.alcor.common.db")
 public class VpcResourceCache {
 
-    // Map <VNI, Map<PIP, List<ResoruceIDType>>
+    // Map <VNI, Map<PIP, ResourceMetadata>
```
pkommoju, line 30: TXN:
Need a "CacheFactory    cacheFactory" member which will be assigned the cacheFactory param passed into the Constructor.

The following method is also needed:
public Transaction getTransaction() {
        return cacheFactory.getTransaction();
}

The same change is needed in HostResourceMetadataCache and, ResourceCacheManager classes.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/server/grpc/GoalStateProvisionerServer.java
```diff
@@ -105,6 +128,17 @@ public void onNext(Goalstate.GoalStateV2 value) {
...
+                    for (Map.Entry<String, HostGoalState> entry : hostGoalStates.entrySet()) {
+                        String hostId = entry.getKey();
+                        HostGoalState hostGoalState = entry.getValue();
+
+                        try {
+                            goalStatePersistenceService.updateGoalState(hostId, hostGoalState);
+                        } catch (Exception e) {
+                            e.printStackTrace();
+                        }
+                    }
```
pkommoju, line 141: TXN:
This entire loop can be pushed into goalStatePersistenceService.updateGoalState() and the call it self invoked after grpcGoalStateClient.sendGoalStates(filteredGoalStates); in the try{} block below.

The abouve loop inside goalStatePersistenceService.updateGoalState()  will have to be modified as follows:
goalStatePersistenceService.updateGoalState() {
try {
Transaction txn = getTransaction();
txn.start();
<The loop body over the current code>
txn.commit();
} cartch () {}

This way, if the send fails we don't have to do any rollbacks.

I have added
public Transaction getTransaction() {
        if(hrmCache != null)
            return hrmCache.getTransaction();
        if (rssCache != null)
            return rssCache.getTransaction();

        if (vpcCache != null)
            return vpcCache.getTransaction();

        return null;
    }

In my equivalent of service/impl/GoalStatePersistenceServiceImpl.java.

There is the issue of capturing changes from updates to ACA into NCM caches, which, I don't think is handled by this code. I might be wrong, in which case a clear comment to that effect will help a lot.

xieus: Sounds good. As discussed, we will merge the current NCM implementation for non-transactional perf test. Meanwhile please work on the other PR on Transaction Mode changes (PR #604) and get it merged. 

Next step is continue refactoring this part to enable transactional support and do a quick perf test.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/service/impl/OnDemandServiceImpl.java
```diff
@@ -0,0 +1,165 @@
...
+        //    been handled by SG label), check if the 5-tuples (ip/port, destination ip/port, protocol) + ethertype
+        //    comply with the outbound SG rules of source port
+        // 3.1 query cache M3 based on associated SG IDs of the source port and retrieve existing SG detail
+        // 3.2 for each SG and SG rule, check if outbound rule allows the 5-tuples + ethertype,
+        //                             => yes, go to step 4 | no, go to Step 3.3
+        // 3.3 (optional) if a rule includes a remote SG id, query cache M3 and retrieve detailed membership of
+        //                remote SG, check whether destination id belongs to the remote SG
+        //                             => yes, go to Step 4 | no, reject
+        // 4. Bingo! this packet is allowed, collect port related resources (NEIGHBOR, SG etc. FULL GS) and send down
+        //    to ACA by a separate gRPC client
```
cj-chung, line 76: @xieus How about routing rule check?

PR #612, opened 2021-05-12
[Schema Change] Add gRPC channel to notify node subscribe MQ

VanderChen:


services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/NodeSubscribeClient.java
```diff
@@ -0,0 +1,26 @@
...
+    WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+package com.futurewei.alcor.dataplane.client;
+
+import com.futurewei.alcor.schema.Subscribeinfoprovisioner;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@Component
```
xieus, line 23: If this is an interface class, we can move the @component annotation to the real implementation NodeSubscribeClientImpl.

schema/proto3/subscribeinfoprovisioner.proto
```diff
@@ -0,0 +1,43 @@
...
+}
+
+message NodeSubscribeInfo {
+  SubscribeOperation subscribe_operation = 1;
+  string topic = 2;
+  string key = 3;
+}
+
+message SubscribeOperationReply {
+  bool is_success = 1;
```
xieus, line 42: For the response, we can use an existing operation response entity: 

https://github.com/futurewei-cloud/alcor/blob/64f5d8315c9c7d1335c4593438e09285f725efce/schema/proto3/common.proto#L43

schema/proto3/subscribeinfoprovisioner.proto
```diff
@@ -0,0 +1,43 @@
...
+
+enum SubscribeOperation {
+  Subscribe = 0;
+  Unsubscribe = 1;
+}
+
+message NodeSubscribeInfo {
+  SubscribeOperation subscribe_operation = 1;
+  string topic = 2;
+  string key = 3;
```
xieus, line 38: What does the key differ from topic?

xieus (changes requested):

schema/proto3/subscribeinfoprovisioner.proto
VanderChen: I will use OperationStatus to replace the bool variable.

schema/proto3/subscribeinfoprovisioner.proto
VanderChen: The key is for key-shared mode in Pulsar. It's similar to the concept of a secondary topic in the message queue (e.g. key in Apache Pulsar, tag in RocketMQ). At the same time, this value can be set to null when it is not needed.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/NodeSubscribeClient.java
VanderChen: I have fixed it.

schema/proto3/subscribeinfoprovisioner.proto
xieus: It would be great to put the comment of "Allow to set NULL when a secondary topic is not needed" for the key field. @VanderChen

xieus (approved):
LGTM. 

@VanderChen I have one last minor comment. Please take a look and if you can fix it in your other PR, that would be great.

PR #550, opened 2021-03-02
[Message Queue] Design Update

issacyxw:

PR #613, opened 2021-05-12
[Data Plane Mgr] Support vpc-mode pulsar message queue

VanderChen:
Implement the design PR #550


VanderChen:
Due to the huge difference in design, I don't think this PR is suitable for merge. How to modify the code to improve the versatility needed further discussion.

PR #618, opened 2021-05-27
[CI/CD] Jenkins test script fix

pkommoju:
Possible fix for jenkins job busybox_ping_test_basic_case_with_build failing.

__Issue Description__
The URL seen by the PM controller has %22 prefix and suffix on the projectid. The projectid retrieved by the test script (ping_test.py) has double quotes always but only PM seems to fail while other services seems to work just fine.

Trying a quick fix by removing double quotes from project_id in get_projectid helper function.


xieus (approved):
LGTM

PR #619, opened 2021-05-27
[Data Plane Mgr] Fix GoalState for L3 neighbors on the same host

cj-chung:


services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/NeighborService.java
```diff
@@ -140,10 +133,6 @@ public void buildNeighborStates(NetworkConfiguration networkConfig, String hostI
                         throw new NeighborInfoNotFound();
                     }
 
-                    if (hostIp.equals(neighborInfo.getHostIp())) {
-                        continue;
-                    }
```
xieus, line 145: Will removal of this block put L2 neighbors on the same host in the GS message?

cj-chung: I have included the checking for NeighborType.L3 if hostIps are the same to bypass the L2 neighbor_state in the same host.

xieus (approved):
LGTM

PR #620, opened 2021-05-28
[CI/CD] Jenkins fix

pkommoju:
Fix for KeyError on device_ids.
Add debug output to dump the data in the change_ports when creat ports
is called from other tests.


xieus (approved):
LGTM.

PR #627, opened 2021-06-05
[Data Plane Mgr] Fix bug in processNeighborConfiguration for test scenario 5.

cj-chung:
Test Scenario5:
1) create a network
2) create two subnets
3) create VMs
4) create a router
5) attach two subnets to the router.


services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/DpmServiceImpl.java
```diff
@@ -188,11 +189,18 @@ private void patchGoalstateForNeighbor(UnicastGoalState unicastGoalState) throws
...
 
-                            // Add subnet to router_state
+                            String routerId = subnetEntity.getRouterId();
+                            if (routerId == null) {
+                                List<InternalRouterInfo> internalRouterInfos = networkConfig.getInternalRouterInfos();
+                                for (InternalRouterInfo internalRouterInfo : internalRouterInfos) {
+                                    routerId = internalRouterInfo.getRouterConfiguration().getId();
+                                    if (routerId != null) break;
+                                }
+                            }
```
xieus, line 199: Does this new block of code try to reset routerId? If so, please put some comment to explain the context.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/DpmServiceImpl.java
```diff
@@ -373,7 +383,15 @@ private void patchGoalstateForNeighbor(UnicastGoalState unicastGoalState) throws
                 //unicastGoalState.setHostIp(neighborInfo.getHostIp());
                 unicastGoalState.setHostIp(hostIp);
                 unicastGoalState.getGoalStateBuilder().addNeighborStates(neighborState);
-                patchGoalstateForNeighbor(unicastGoalState);
+                UnicastGoalState unicastGoalStateTemp = new UnicastGoalState();
+                unicastGoalStateTemp.getGoalStateBuilder().addNeighborStates(neighborState);
+                patchGoalstateForNeighbor(networkConfig, unicastGoalStateTemp);
```
xieus, line 388: This block of code appears to be very customized. Was a UnicasetGoalStateTemp 100% necessary? If it would, let us put comment to explain the context.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/DpmServiceImpl.java
cj-chung: I have put comments in the code. 
Basically, this function will be called by two test scenario (#4 and #5). 
Test scenario #4 has routerId inside of subnet, but test scenario #5 doesn't, so test scenario #5 needs get routerId from networkCofnig.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/DpmServiceImpl.java
cj-chung: I have put comments in the code.
`unicastGaolStateTemp` is to get patchGoalStates (subnet_states and a consolidated router_state) based on the current neighborState.

xieus: Thanks James. Is UnicastGoalStateTemp reusable? If it could be created once and reused by multiple entry in the for loop, it could save us something.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/DpmServiceImpl.java
```diff
@@ -373,7 +385,19 @@ private void patchGoalstateForNeighbor(UnicastGoalState unicastGoalState) throws
...
+
+                unicastGoalState.getGoalStateBuilder().addAllSubnetStates(unicastGoalStateTemp.getGoalStateBuilder().getSubnetStatesList());
+                unicastGoalState.getGoalStateBuilder().addAllRouterStates(unicastGoalStateTemp.getGoalStateBuilder().getRouterStatesList());
+                unicastGoalState.getGoalStateBuilder().addAllSubnetStates(multicastGoalState.getGoalStateBuilder().getSubnetStatesList());
+                rebuildRouterState(unicastGoalState.getGoalStateBuilder(), multicastGoalState.getGoalStateBuilder());
+                multicastGoalState.getGoalStateBuilder().addAllSubnetStates(unicastGoalStateTemp.getGoalStateBuilder().getSubnetStatesList());
+                rebuildRouterState(multicastGoalState.getGoalStateBuilder(), unicastGoalStateTemp.getGoalStateBuilder());
```
xieus, line 400: Let us see if line 391-400 could be patched in a standalone method.

cj-chung: These lines of code are very depends on the scenario, it's only for the test scenario #5, to patch neghbor_state and rebuild router_state. Unless we have new scenario to use the same sequence of patching (I cannot think of any now), otherwise it's not worth to create a method especially for this case.

cj-chung: The unicastGoalStateTemp is a local variable, that's is a variable to get the output of the `patchGoalstateForNeighbor`. Since I made this function for both scenario #4 and #5, so here the unicastGoalStateTemp is necessary to get patched states, otherwise if we use original unicastGoalState or multicaseGoalState, it will mess up both states.

xieus: Makes sense. Thanks.

xieus (approved):
LGTM

PR #637, opened 2021-06-10
[Communication] gRPC channel/stub pool and keepalive for channels

zzxgzgz:
This PR does the following:

1. Changes the `GoalStateClientImpl` to singleton model.
2. Implemneted gPRC channel and stub pool in `GoalStateClientImpl`, so that one host IP uses the same channel and stub, if they are healthy.
3. Implemented logic to check gRPC channel status and reconnection(if needed), before sending GoalState messages.

Comparison:
image
This is a few test result I did for comparing the NCM on master branch(without channel/stub pooling), and with different versions of NCM that I implemented on this branch. 

We can see from column I(Ping Speed(ms) Avg.) that, without any pooling, or even with channel pool(but not stub pool), which is from testcase 8 and 9, the average ping speed is at about ~400ms.

After adding stub pooling, the average ping speed increased to ~200, ~300ms.

### DIFF
--- a/services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
+++ b/services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
@@ -47,7 +47,7 @@ public class GoalStateClientImpl implements GoalStateClient {
 
     private final ExecutorService executor;
 
-    private SortedMap<String, GrpcChannelStub> hostIpGrpcChannelStubMap;
+    private ConcurrentHashMap<String, GrpcChannelStub> hostIpGrpcChannelStubMap;
 
 
     public static GoalStateClientImpl getInstance(){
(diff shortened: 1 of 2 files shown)


xieus:
@zzxgzgz I will catch up on this PR review.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
```diff
@@ -61,6 +75,7 @@ public GoalStateClientImpl() {
                 new DefaultThreadFactory("grpc-thread-pool"));
 
         //TODO: Setup a connection pool. one ACA, one client.
+        this.hostIpGrpcChannelStubMap = new TreeMap<>();
```
xieus, line 78: TreeMap is an implementation that is not synchronized (Ref: https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html)

Do you consider ConcurrentHashMap?

zzxgzgz: > TreeMap is an implementation that is not synchronized (Ref: https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html)
> 
> Do you consider ConcurrentHashMap?

Changed this map to ConcurrentHashMap.

xieus (approved):
LGTM!

PR #642, opened 2021-06-14
[API Perf Enhancement] Optimize vpc create default segment table

DavidLiu506:
Create default table optimize from about 10 second to about 1 second.


xieus (approved):
LGTM. @DavidLiu506 Congratulations to the first commit.

PR #661, opened 2021-07-08
[Subnet Mgr] Fix gateway port duplicated creation issue

xieus:
This PR has the following changes:
- Fix gateway port duplicated creation issue when a subnet is created
- Remove legacy codes in Subnet Manager


cj-chung (approved):
LGTM

PR #670, opened 2021-07-19
Warm up gRPC channels and add log for NCM performance analysis

zzxgzgz:
This PR does the following: 

1. Added logs in NCM, so that it can be used for performance analysis, along with ACA's log and Test Controller's log.
2. Added code for NCM's gRPC channels so that users can configure how many gRPC channels NCM establishes to each ACA, also how many warmup GoalStates each channel will send when it is first instantiated.

Notes:
1. To enable the logging for performance, please change the `logging.level.root` in `services/network_config_manager/src/main/resources/application.properties` to `DEBUG`, the `DEBUG` level equals the `Level.FINE` in the java code.
2. When a coder sets some values in `application.properties`, in order to retrive the values correctly, you need to retrieve them in a class that has either `@Component` or `@Bean`, otherwise, it doesn't work and the value is always the default value of that class, such as 0 for an `int`, or null for other Objects.


zzxgzgz:
Adding a sample output of the analysis for your reference:

```
$ python3 analyze.py ./aca.log ./ncm_log_x.log ./tc.log 
Hi
Argument: analyze.py
Argument: ./aca.log
Argument: ./ncm_log_x.log
Argument: ./tc.log
This file has 26497 lines
For keyword [Elapsed time for update goalstate operation took:], there are 33 lines
For On-demand GS process time, the min is 0.0 ms, max is 60.0 ms, mean is 19.000 ms, median is 21.0
For keyword [T3], there are 20 lines
For On-demand call send to receive time (T3 - T1), the min is 92.0 ms, max is 173.0 ms, mean is 146.789 ms, median is 148.0
For keyword [[METRICS] Elapsed time for port operation took:], there are 33 lines
For Port GS Process Time, the min is 0.0 ms, max is 59.0 ms, mean is 7.606 ms, median is 8.0
For keyword [[METRICS] Elapsed time for neighbor operation took:], there are 33 lines
For Neighbor GS Process Time, the min is 0.0 ms, max is 49.0 ms, mean is 9.788 ms, median is 10.0
This file has 8794 lines
For keyword [From received hostOperation to before response], there are 20 lines
For On-demand call received to response sent time (~T3 - T2), the min is 60.0 ms, max is 125.0 ms, mean is 108.850 ms, median is 111.0
For keyword [retrieved vpc resource metadata, elapsed Time in milli seconds:], there are 20 lines
For Retrieve VPC resource metadata time, the min is 5.0 ms, max is 20.0 ms, mean is 16.050 ms, median is 17.0
This file has 2977 lines, total pings: 20, successful_pings: 20, failed pings: 0
For keyword [bytes from], there are 20 lines
For Ping Speed, the min is 95.865 ms, max is 188.29 ms, mean is 155.183 ms, median is 158.5075
Ping speed is as follows:
150.454
95.865
169.558
167.035
128.88
128.794
151.262
175.191
170.386
173.43
131.117
171.236
188.29
183.667
147.877
171.802
165.753
148.696
137.401
146.969
Done printing ping speed
This ACA log file has 20 neighbor IDs
This ACA log file has 20 on-demand call UUIDs
For host operation request grpc delay, the min is 0.0 ms, max is 25.0 ms, mean is 12.150 ms, median is 10.0
For host operation reply grpc delay, the min is 22.0 ms, max is 30.0 ms, mean is 25.100 ms, median is 24.0
There are totally 20 unique neighbor IDs, and 20 of them appear both in ACA and NCM log
For ncm push goalstate grpc delay, the min is 2.0 ms, max is 21.0 ms, mean is 7.400 ms, median is 3.5
For ACA received ncm reply to packet out time, the min is 1.0 ms, max is 15.0 ms, mean is 6.150 ms, median is 4.5
```

services/network_config_manager/src/main/resources/application.properties
```diff
@@ -24,6 +24,8 @@ dataplane.isovs=true
 grpc.min-threads = 100
 grpc.max-threads = 200
 grpc.threads-pool-name = grpc-thread-pool
+grpc.number-of-channels-per-host = 10
+grpc.number-of-warmups-per-channel = 1
```
xieus, line 28: What would happen if admin misconfigures these two channel numbers, say 5 and 10 :-)

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/service/impl/OnDemandServiceImpl.java
```diff
@@ -84,31 +84,38 @@ public HostGoalState retrieveGoalState(Goalstateprovisioner.HostRequest.Resource
...
-
+        long start = System.currentTimeMillis();
 //        ResourceMeta resourceMetadata = retrieveResourceMeta(vni, sourceIp);
         VpcResourceMeta vpcResourceMetadata = retrieveResourceMeta(vni);
         if (vpcResourceMetadata == null) {
             logger.log(Level.INFO, "[retrieveGoalState] retrieved vpc resource metadata is null | vni = " + vni);
             return null;
         }
-
+        long end = System.currentTimeMillis();
```
xieus, line 94: We would need an empty line after "}" (Line 93).

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
```diff
@@ -170,17 +231,26 @@ public void onCompleted() {
...
         StreamObserver<Goalstate.GoalStateV2> requestObserver = asyncStub.pushGoalStatesStream(responseObserver);
         try {
-                Goalstate.GoalStateV2 goalState = hostGoalState.getGoalState();
-                logger.log(Level.INFO, "Sending GS to Host " + hostIp + " as follows | " + goalState.toString());
-
-                requestObserver.onNext(goalState);
+            Goalstate.GoalStateV2 goalState = hostGoalState.getGoalState();
+            logger.log(Level.INFO, "Sending GS to Host " + hostIp + " as follows | " + goalState.toString());
+            requestObserver.onNext(goalState);
+            if (hostGoalState.getGoalState().getNeighborStatesCount() == 1 && (hostIp.equals("192.168.20.92") || hostIp.equals("10.213.43.92"))) {
```
xieus, line 237: This is internal test IPs and we should get rid of them all in the master codes.

xieus (commented):
@zzxgzgz Pls take a look at a few comments.

services/network_config_manager/src/main/resources/application.properties
zzxgzgz: I think, as long as they are positive numbers, then they should be good. 5 channels per host and 10 warmups in each channel could be desirable for some users. But anyway, maybe I need to check for 0 and negative numbers, let me add the code for that.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
```diff
@@ -47,35 +48,36 @@
 
     private final ExecutorService executor;
 
-    private ConcurrentHashMap<String, GrpcChannelStub> hostIpGrpcChannelStubMap;
+//    // each host_ip should have this amount of gRPC channels
```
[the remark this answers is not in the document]
zzxgzgz: Fixed

services/network_config_manager/src/main/resources/application.properties
zzxgzgz: Added sanity check for these two variables

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/service/impl/OnDemandServiceImpl.java
zzxgzgz: Added

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
zzxgzgz: Removed those IPs from the code, and created a varaible called `monitorHosts` in the java code, which comes from a variable called `grpc.monitor-hosts` in `application.properties`.

services/network_config_manager/src/main/resources/application.properties
```diff
@@ -24,6 +24,9 @@ dataplane.isovs=true
...
+grpc.monitor-hosts = 0.0.0.0,0.0.0.1
```
xieus, line 29: @zzxgzgz if we add a new compute node, do we have to update this configuration? If so, this is basically all the nodes this NCM instance needs to monitor. 

We could leverage the Node Cache inside NCM, which has the requested information already.

xieus (approved):

services/network_config_manager/src/main/resources/application.properties
zzxgzgz: Didn't know that we have a node cache in NCM. 

However, in my tests, I am only monitoring one host, the different IP addresses is just the different interfaces of the same host. I think it is easy for us to edit.

PR #671, opened 2021-07-21
[Data Plane Mgr] Allow empty routing table when creating router state

DavidLiu506:
This PR is to address an aggression issue that causes L3 neighbors unable to ping each other.


cj-chung (approved):
LGTM

PR #678, opened 2021-08-13
[Jaeger] Fix tracing spans between functions and services in PM

cj-chung:
Fix tracing spans across concurrent async calls in PM


services/port_manager/src/main/java/com/futurewei/alcor/portmanager/request/RequestManager.java
```diff
@@ -58,16 +64,28 @@ public void sendRequest(IRestRequest request) throws Exception {
...
+                        sendRequest(request, callback);
+                    } catch (Exception e) {
+                        throw new CompletionException(e);
+                    }
+                    return null;
+                }
+            }, AsyncExecutor.executor).thenRun(span::finish);
+
+            addFuture(request, future);
+        }
```
xieus, line 88: Do we need catch for the external try block?

cj-chung: Yes. We can remove the outer try-catch.

xieus (approved):
LGTM

PR #680, opened 2021-08-20
[Subnet Mgr] Allow to delete a subnet without a gateway port

yanmo96:
This is a short-term fix to the following Issue.

__Issue__

When subnet's gateway port (device_owner=router_interface) got deleted, subnet's GatewayPortId still has the port ID (PM delete the port but didn't notify SM). 

__Short-term Fix__

We temporarily silence the exception when deleteGatewayPort throws an exception. An more appropriate fix should be on port manager side to notify subnet manager.


services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/controller/SubnetController.java
```diff
@@ -497,7 +497,11 @@ public ResponseId deleteSubnetState(@PathVariable String projectId, @PathVariabl
             // TODO: delete gateway port in port manager. Temporary solution, need PM fix issue
             GatewayPortDetail gatewayPortDetail = subnetEntity.getGatewayPortDetail();
             if (gatewayPortDetail != null) {
-                this.subnetToPortManagerService.deleteGatewayPort(projectId, gatewayPortDetail.getGatewayPortId());
+                try{
+                    this.subnetToPortManagerService.deleteGatewayPort(projectId, gatewayPortDetail.getGatewayPortId());
+                } catch (HttpClientErrorException.NotFound e) {
+                    logger.warn(e.getMessage());
+                }
```
xieus, line 504: We don't want to completely silence this expecting. In some case the exception may be valid. 

@yanmo96  what is the portId returned from gatewayPortDetail.getGatewayPortId() when the exception was thrown? Is it 
Null or empty?

Two options: 
(1) we don't need to call deleteGatewayPort as there is no port to be deleted. 
(2) let this special handling as part of deleteGatewayPort method. 

Thanks.

cj-chung: I think the problem is from the subnet's gateway port (device_owner=router_interface) got deleted, and subnet's GatewayPortId still has the port ID (PM delete the port but didn't notify SM).

xieus: I see. Then this is a workaround then. In this case, let us mark this fix as a short-term fix. 

@cj-chung we may want to create an issue for tracking purpose. I will merge current fix.

cj-chung: Issue is created: https://github.com/futurewei-cloud/alcor/issues/681

xieus:
Linked to issue #681

xieus (approved):

Issue #682, opened 2021-08-27
[PM] PM Failed to delete port when concurrent


yanmo96:
Test case: alcor-create-and-delete-ports
- create and delete 10 ports per network, do it 2 times with 2 concurrency.

Issue:
- During clean up it will show there are still ports in the subnet, so subnet can not be deleted.

Root cause:
- Two process are accessing the cache at same time, and one process put back the port already been by the other process.

`
2021-08-25 11:01:40.827 26006 DEBUG neutronclient.v2_0.client [-] Error message: {"timestamp":"2021-08-25T18:01:40.826+0000","status":409,"error":"Conflict","message":"There is some ports in the subnet, we can Not delete subnet","path":"/project/05fd124b97484cb8963a58a8f1b6a268/subnets/ee9e889e-9be4-437f-95c9-ccf225b6e87c"} _handle_fault_response /root/rally_venv/lib/python3.6/site-packages/neutronclient/v2_0/client.py:262
`

PR #683, opened 2021-08-27
Change the atomicity mode of some caches to transactional to fix issue 682

pkommoju:
Change all caches in Port Manager to transactional atomicity. Mixing caches of different atomicity mode under a transaction does not work in Ignite. This was  causing a failure in concurrent deletion of ports.

Linked to Issue #682


xieus:
How about this PR? @pkommoju We are seeing quite a few of conflicting files.

pkommoju:
Yes, this is expected because changes in two branches are getting combined into one. Let me take a look at them and attach a patch/comments about what to keep what to remove.

pkommoju:
Abandon this PR and branch. I didn't realize it quickly enough.

PR#685 already brought in all these changes.

xieus:
Close this PR as it is duplicated with PR #685.

PR #687, opened 2021-09-09
[CI/CD] Automate building ACA from jenkins jobs

pkommoju:
* As part of build automation ACA is also built from latest sources along with Alcor.
* There is provision for specifiying a repository, branch and a commit to validate ACA code changes from within Jenkins before merging into master. There is also a provision to force a new build regardless of source changes.
* ACA is built parallely on all listed ACA nodes first, before building Aclor. If building ACA failes, the job itself will fail.
* ACA builds take quite longtime because a lot of 3rd party libraries are built from sources each time. This can be changed so that they are built only if there are changes.


scripts/busybox-ping-test/build_aca.sh
```diff
@@ -0,0 +1,160 @@
+#! /bin/sh
+
+# Pull latest ACA code from specified repository, branch and commit
```
xieus, line 3: A general comment: Having a script in Alcor repo for ACA build sounds a bit awkward. Let us plan to migrate those scripts to Alcor-Int repo so that they are independent of component repos like Alcor and ACA.

scripts/busybox-ping-test/build_aca.sh
```diff
@@ -0,0 +1,160 @@
...
+
+
+# Get remote repo/branch/commit info
+# May need git fetch -all to be robust
+git_check_remote() {
+    REMOTE_INFO=`git log origin/$GIT_BRANCH | head -3`
+    REMOTE_COMMIT=`echo $REMOTE_INFO | awk '/^commit/ {print $2}'`
+    REMOTE_DATE=`echo $REMOTE_INFO |  sed 's/^.*Date://' | awk '{print $1, $2, $3, $4, $5}'`
+
+    RSEC=`date --date="${REMOTE_DATE}" +%s`
```
xieus, line 33: Same date might have multiple commits. What do we do with that? :-)

scripts/busybox-ping-test/remote_aca_build.sh
```diff
@@ -0,0 +1,86 @@
...
+for node in `echo ${ACA_NODES}`; do
+    rm -f /tmp/aca_${node}_build.log > /dev/null 2>&1
+    NC=`expr $NC + 1`
+    scp ./build_aca.sh ubuntu@${node}:${ACA_DIR}/
+    ssh ubuntu@$node "cd $ACA_DIR && sudo ./build_aca.sh $ACA_REPO $ACA_BRANCH $ACA_COMMIT $DO_FORCE" > /tmp/aca_${node}_build.log 2>&1 &
+    echo "Scheduled build on $node"
+done
+
+# check the status
+SC=0
```
xieus, line 69: Do we have a dummy API in ACA for us to verify it is up?

pkommoju: No. one or more of ps output, netstat output, and possibly trawling through syslog allow us to learn if ACA is running or not.

scripts/busybox-ping-test/build_aca.sh
pkommoju: Full dates are converted into seconds since epoch before comparing. It would not be possible for two commits have same timestamp down to the last second since each commit effectively locks the repository until it is done. So we are fine.

scripts/busybox-ping-test/build_aca.sh
pkommoju: I thought about considerably and made this decision because the current design of the Jenkins job used to run CI doesn't allow more than one repository.

Yes, sometime in near future it can be replaced with "pipeline" and other designs which allow working with multiple repositories.

xieus: Think about this: Consolidate all scripts in one place, like in Alcor-int, which uses git sub-module (https://git-scm.com/book/en/v2/Git-Tools-Submodules) to point to multiple repos including Alcor controller, ACA and others. We don't need to make the move now but time to plan it :-)

scripts/busybox-ping-test/build_aca.sh
xieus: Cool that is what we need exactly.

xieus (approved):
LGTM. One additional comment.

scripts/busybox-ping-test/build_aca.sh
pkommoju: I will keep this in mind.

PR #700, opened 2021-11-02
Attach router bug workaround in S5.

pkommoju:
Temporary workaround for S5 failing.
When subnet contains host routes, route manager requires the subnet to
contain a router and raiese error when that a router is being attached
to that subnet.
This tempoary fix removes host_routes from subnet info while actual fix
is being worked on.


xieus:
Discussed with @pkommoju and @DavidLiu506. As this PR is a temporary fix, we are safe to close it as the actual fix has been merged to master.

PR #701, opened 2021-11-03
Fix subnet attachment to router when subnet has host routes

DavidLiu506:
Attach subnet to router, if there are host routes, subnet manager will tell router manager route table but didn't send routerid to router manager. Router manager can not get this subnet attached routerid because subnet still not complete attachment.
Send routerid to router manger works.


services/subnet_manager/src/main/java/com/futurewei/alcor/subnet/service/implement/SubnetServiceImp.java
```diff
@@ -487,7 +487,7 @@ public void updateSubnetRoutingRuleInRM(String projectId, String subnetId, Subne
         routetable.setRouteTableType(RouteTableType.NEUTRON_SUBNET.getRouteTableType());
         routetable.setRouteEntities(routeEntities);
 
-        String routeManagerServiceUrl = routeUrl + "project/" + projectId + "/subnets/" + subnetId + "/routetable";
+        String routeManagerServiceUrl = routeUrl + "project/" + projectId + "/subnets/" + subnetId + "/routetable?routerid=" + subnetEntity.getAttachedRouterId();
```
xieus, line 490: Do we always preassign value to SubnetEntity's attachedRouterId?

DavidLiu506: Yes.

DavidLiu506: Yes, this is for host route.

xieus (approved):

PR #703, opened 2021-11-19
[DB] Support SQL Field Query in Alcor

pkommoju:
Ignite SqlFieldQuery changes.

* Ignite allows fast searching of the KV store using a field (secondary key/index) in the value field through SQL interface.
* This change is almost entirely in the commonlib (Ignite interface layer).
* To enable this feature, add the annotation @QuerySqlField(index = true) to the field to index on.
* When a class member is annotated with  @QuerySqlField(index = true), a get() using filterParams will use SQLFieldsQuery if all the filter params are also in the set of  QuerySqlFields.
* If fields m1, m2, and m3 are in the  QuerySqlField set, in that order, and thet get supplies only m1, or m1 and m2, or all three, then Ignite interface will use SqlFieldsQuery, otherwise it will fallback to Scan Query. This is called index prefix search.
* One optimization not in this commit is to cache the queries and reuse them. If needed, this optimization can be added later.
* For an example use, look at web...NodeInfo class. and services/sqlquery and services/scanquery.


lib/src/main/java/com/futurewei/alcor/common/entity/CustomerResource.java
```diff
@@ -17,12 +17,14 @@
...
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.ignite.cache.query.annotations.QuerySqlField;
 
 import java.util.Objects;
 
 public class CustomerResource extends Resource {
 
     @JsonProperty("project_id")
+    // @QuerySqlField(index = true)
```
cj-chung, line 27: @pkommoju If we don't need add `@QuerySqlField` here, please remove it and the import.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/NetworkConfigManagerApplication.java
```diff
@@ -32,18 +32,23 @@
...
+         * Return so that rest API endpoint can work
+         *
         try {
             networkConfigServer.start();
             networkConfigServer.blockUntilShutdown();
         } catch (Exception e) {
             e.printStackTrace();
         }
     }
+        */
```
cj-chung, line 51: @pkommoju Why you comment out this part of code?

cj-chung: Should we start grpcServer here?

lib/src/main/java/com/futurewei/alcor/common/entity/CustomerResource.java
cj-chung: @DavidLiu506 Please confirm, if we need the index on `project_id`?

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
```diff
@@ -32,9 +33,11 @@
     private static final Logger logger = LoggerFactory.getLogger(NodeInfo.class);
 
     @JsonProperty("node_id")
+    // @QuerySqlField(index = true)
     private String id;
 
     @JsonProperty("node_name")
+    // @QuerySqlField(index = true)
```
cj-chung, line 40: @pkommoju and @DavidLiu506 Do we need index for `node_id` and `node_name`? If not, please remove it.

lib/src/main/java/com/futurewei/alcor/common/entity/CustomerResource.java
pkommoju: I will remove it after Dahai confirms that it is no longer needed.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/NetworkConfigManagerApplication.java
pkommoju: That comment should be removed, I will remove it.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
pkommoju: node_name needs index. In fact NMM swagger and the getAllNodes allows querying on any combination of name, mac, local ip, ncm_id. I should add index on all of them. id may not be needed if getAllNodes doesn't use SCAN query if querying by only id. I will check this path and remove index on id if this is so.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/IgniteClientDbCache.java
```diff
@@ -223,4 +283,209 @@ public long size() {
...
+        }
+        catch (Exception e) {
+            logger.log(Level.WARNING, "runSQLFieldsQuery exception : " + e.getMessage());
+        }
+
+        return null;
+    }
+
+    @DurationStatistics
+    public <E1, E2> Map<K, V> getBySqlFieldsAll(Map<String, Object[]> filterParams) {
```
cj-chung, line 328: @pkommoju Do we need to throw exception from this function?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/IgniteClientDbCache.java
```diff
@@ -37,66 +44,113 @@
...
+            clientCacheConfig.setName(CommonUtil.getSqlNameFromCacheName(cacheConfig.getName()));
             clientCacheConfig.setAtomicityMode(cacheConfig.getAtomicityMode());
             logger.log(Level.INFO, "Getting or creating cache " + clientCacheConfig.getName() + " AtomicityMode is " + clientCacheConfig.getAtomicityMode());
-            this.cache = igniteClient.getOrCreateCache(clientCacheConfig);
-            logger.log(Level.INFO, "Retrieved cache " +  this.cache.getConfiguration().getName() + " AtomicityMode is " + this.cache.getConfiguration().getAtomicityMode());
+            extractSqlFields(className);
+            if (sqlFields != null && sqlFields.size() != 0) {
+                this.cache = getOrCreateIndexedCache(igniteClient, className, clientCacheConfig, null);
+                if (this.cache == null) {
+                    logger.log(Level.WARNING, "Create cache for client " + className + " with index failed, falling back");
```
cj-chung, line 109: @pkommoju How to handle falling back if `this.cache=null` ? just use `igniteClient.getOrCreateCache(clientCacheConfig)` instead here?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/IgniteClientDbCache.java
pkommoju: I think there is no need. runSqlFieldsQuery will throw any exception anyway. Just need to add throws CacheException.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/IgniteClientDbCache.java
pkommoju: I am logging it as a warning message here. Few lines below, if this.cache is null, we call igniteClient.getOrCreateCache(clientCacheConfig). This call could also be because the the cache is not setup for SQLFieldsQuery in addition to handling the case where it is setup but something went wrong. Didn't want to have the call in two places.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/NetworkConfigManagerApplication.java
cj-chung: Let's just leave comment here. If we need it in the future, we can enable it then.

pkommoju: This is a temporary change, it should have been removed but permanent change is coming from my other PR about NMM to NCM by REST. I will leave an exploratory comment disabling this section in there but leave this code in this PR as is.

xieus:
@pkommoju There is a conflict with existing master after PR #704 was merged.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/NetworkConfigManagerApplication.java
pkommoju: I think I will remove the comment here since the other PR will have a proper fix and SQL fields query doesn't strictly need NCM.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
pkommoju: Removed index on id, added it on localIp, macAddress, ncm_id.

lib/src/main/java/com/futurewei/alcor/common/entity/CustomerResource.java
pkommoju: Removed.

docs/modules/ROOT/pages/performance/sql_and_scan_query_pref_comp.adoc
```diff
@@ -0,0 +1,48 @@
...
+In Alcor there are some cases where queries are made on the cases using a field other than the key. Since Ignite is the KV database, these types of queries will require full scan (linear search) of the cache in question and this introduces big latencies as the number of entries in the cache grows.
+
+Ignite supports secondary keys, called QuerySqlFields which will use a tradional B+Tree indexes which can answer point lookup and also range, minumum, maximum, greater than, and less than queries mush faster.
+
+This report is about comparing the performance of the queries on non-key fields using Ignite's SCAN query and through SQL queries.
+
+== Test setup
+Operations are on NodeInfo object. Benchmark tool, DPM, NCM and NMM run on 10.213.43.161.
+Ignite is ru on 10.213.43.163 and 10.213.43.164.
+Number of entries in both SQL and Scan are 1M, 1000 entries are queried on.
```
cj-chung, line 18: @pkommoju I think it's not good to show ip addresses here.

pkommoju: removed IP addresses.

cj-chung (changes requested):
@pkommoju Another comment: please rename the `scanquery` service to a name like `scanquery_test_nodemanager`. 

In that case, should the sqlquery service also be renamed as sqlquery_test_nodemanager, to be consistant?

lib/src/main/java/com/futurewei/alcor/common/db/ignite/IgniteClientDbCache.java
```diff
@@ -37,66 +44,113 @@
...
+    private static final String SQL_SCHEMA_NAME = "alcor";
     private ClientCache<K, V> cache;
     private final IgniteClientTransaction transaction;
+    private static class SqlField {
+        public String type;
+        public boolean isIndexed;
+    }
+    private Map<String, SqlField> sqlFields = null; // needed for index creation and querying
+    private List<String> sqlColumns = null;         // needed for constructing indexes in specified order
+                                                    // so that a prefix query can use indexed search.
```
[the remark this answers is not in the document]
pkommoju: Removed sqlColumns and changed Map to LinkedHashMap.

lib/src/main/java/com/futurewei/alcor/common/db/ignite/IgniteClientDbCache.java
```diff
@@ -37,66 +44,113 @@
...
             clientCacheConfig.setAtomicityMode(cacheConfig.getAtomicityMode());
             logger.log(Level.INFO, "Getting or creating cache " + clientCacheConfig.getName() + " AtomicityMode is " + clientCacheConfig.getAtomicityMode());
-            this.cache = igniteClient.getOrCreateCache(clientCacheConfig);
-            logger.log(Level.INFO, "Retrieved cache " +  this.cache.getConfiguration().getName() + " AtomicityMode is " + this.cache.getConfiguration().getAtomicityMode());
+            extractSqlFields(className);
+            if (sqlFields != null && sqlFields.size() != 0) {
+                this.cache = getOrCreateIndexedCache(igniteClient, className, clientCacheConfig, null);
+                if (this.cache == null) {
+                    logger.log(Level.WARNING, "Create cache for client " + className + " with index failed, falling back");
+                }
```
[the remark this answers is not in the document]
pkommoju: This check is a few lines above.

DavidLiu506 (approved):

cj-chung (approved):
LGTM

PR #704, opened 2021-11-23
Send GoalStateV2 from DPM, NCM and ACA

DavidLiu506:
1 Merge GoalState in DPM and send to NCM, wait for NCM response.
2 Send GoalState from NCM to ACA and wait for all ACA response.
3 Improve NCM performance.
4 Use linkerd load balance grpc request for NCM.

PR #707, opened 2021-12-01
[Tracing] Support Jaeger tracing in NCM and TC

### LINKED ISSUE #708: [NCM] NCM Not Putting VPC States into Cache during PushGoalStatesStream
In
https://github.com/futurewei-cloud/alcor/blob/master/services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/service/impl/GoalStatePersistenceServiceImpl.java#L64

NCM didn't put VPC states into cache, which leads to retrieving a `null` VPC state, given a valid `VPC_ID`, during the on-demand workflow.

zzxgzgz:
What does this PR do:

1. Added Jaeger tracing to NCM's gRPC server/client, so that we can use Jaeger to trace NCM's gRPC activities.
2. Added Jaeger tracing to the Test Controller, to make the tracing more complete.
3. Closes #708 

Please refer to the following screenshot to get an idea of what to expect, when running Test Controller:
GoalState Pushdown process:
image

On-demand process:
image
The on-demand tracing was captured before merging #704 , after merging this PR, on-demand doesn't get triggered, as neigbhor states all push down before the pings.

Future improvements:
1. Improve how the spans are organized when on-demand is triggered, the RequestGoalStates spans show be at the same level, but right now they are chained together. For the reason behind it and a possible solution, please refer to [this issue](https://github.com/opentracing-contrib/java-grpc/issues/59) for a possible solution.
2. Jaeger tracing is also implemented in ACA, but the spans between ACA and NCM doesn't get connected, so the ACA's spans look like this:
image
we suspect that this is because we are using different libraries here. In ACA, jaeger tracing was adding using the opentelemetry-cpp library, however, NCM uses the opentracing library. We suspect this mismatch is causing the ACA spans not connected with the NCM spans. We may want to change to opentelemetry library in Alcor later.


xieus:
@cj-chung Could you review this PR?

cj-chung:
> @cj-chung Could you review this PR?

sure

services/network_config_manager/pom.xml
```diff
@@ -49,11 +49,27 @@ Copyright(c) 2020 Futurewei Cloud
     </repositories>
 
     <dependencies>
```
xieus, line 51: @zzxgzgz A new file called executionHistory.lock is added by this PR. This may be a local file. 

Please check and remove it if so.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
```diff
@@ -244,10 +302,18 @@ public void onCompleted() {
         };
 
         StreamObserver<Goalstate.GoalStateV2> requestObserver = asyncStub.pushGoalStatesStream(responseObserver);
+        long requestObserverEstablished = System.currentTimeMillis();
+        logger.log(Level.FINE, "[doSendGoalState] Established RequestObserver, elapsed Time after stub established in milli seconds: " + (requestObserverEstablished - stub_established));
```
xieus, line 306: This logging is quite extensive. Do we have an easy to turn it off when it is not needed..

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/server/grpc/GoalStateProvisionerServer.java
```diff
@@ -73,14 +88,26 @@
     private GoalStateClient grpcGoalStateClient;
 
     public GoalStateProvisionerServer() {
+        System.setProperty("JAEGER_SERVICE_NAME","alcor-ncm");
```
xieus, line 91: "alcor-ncm" should be in the config file so that we could check in a centralized place.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/server/grpc/GoalStateProvisionerServer.java
```diff
@@ -144,9 +171,24 @@ public void blockUntilShutdown() throws InterruptedException {
...
+                        span = tracer.buildSpan("alcor-ncm-server-pushdown-gs").start();
+                    }
+                    logger.log(Level.INFO, "[pushGoalStatesStream] Got parent span: "+(null == pSpan? "null" : pSpan.toString()));
+                    logger.log(Level.INFO, "[pushGoalStatesStream] Built child span: "+span.toString());
+
+                    Scope cscope = tracer.scopeManager().activate(span);
 
                     logger.log(Level.INFO, "pushGoalStatesStream : receiving GS V2 message " + value.toString());
                     long start = System.currentTimeMillis();
+                    Span storeGsSpan = tracer.buildSpan("alcor-ncm-server-store-gs").asChildOf(span.context()).start();
```
xieus, line 189: same here. The span name should be in a project or service config file, and do a reference here.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/server/grpc/GoalStateProvisionerServer.java
xieus: Oh later I found that that it is actually set by opentracing.jaeger.service-name in the application.properties. Why not use the configuration here?

xieus (commented):
A few comments.

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/client/gRPC/GoalStateClientImpl.java
zzxgzgz: I can change these logs' level to FINE, and if we specify the log level to INFO in the `application.properties` then these logs should not show.

services/network_config_manager/pom.xml
zzxgzgz: Removed

services/network_config_manager/src/main/java/com/futurewei/alcor/netwconfigmanager/server/grpc/GoalStateProvisionerServer.java
zzxgzgz: This one is actually tricky, we actually have defined that in the `application.properties`, however, when the code is trying to retrieve this value, it always gets an empty String or just null. That's why I used `System.setProperty` to set this value.

xieus: This could be the spring bean wiring issue. Maybe create an issue to track it, or use //TODO. It is easy to forget :-)

zzxgzgz: Got it, let me add a //TODO in this code then.

xieus (approved):
LGTM.

PR #709, opened 2021-12-04
[CI/CD] Improve handling of Jenkins error and exit

pkommoju:
* Fix the way ping test deals with errorr internally and returns exit code to the ENV so that Jenkins doesn't report success when there is a failure.


xieus (approved):
LGTM

PR #731, opened 2022-01-11
Update GoalStateClientImpl.java

yanmo96:
Set "Sending GS with size" to not print on INFO level


yanmo96:
All change has been added to @DavidLiu506 's new PR

PR #733, opened 2022-01-18
[Deployment] K8s deployment enhancement

yanmo96:
- Rather than wait certain time for k8s to be active, now the deployment script will check the status of k8s pod to make sure they are all up and running before continue. Thus can tolerate different deployment situations.
- Fixed some url links for ignite config files.

### DIFF
--- a/kubernetes/createYaml.sh
+++ b/kubernetes/createYaml.sh
@@ -28,32 +28,36 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
   
   echo "Create yaml files done"
   
-  sleep 40s
-  
+  numberOfRnningPods="foo"
+  numberOfAllPods="bar"
+
+  while [ "$numberOfRnningPods" != "$numberOfAllPods" ];
+  do
+    sleep 10s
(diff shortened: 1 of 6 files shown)


kubernetes/createYaml.sh
```diff
@@ -28,32 +28,36 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
...
   echo "Create yaml files done"
   
-  sleep 40s
-  
+  numberOfRnningPods="foo"
+  numberOfAllPods="bar"
+
+  while [ "$numberOfRnningPods" != "$numberOfAllPods" ];
+  do
+    sleep 10s
```
cj-chung, line 36: @yanmo96 Why we still need to wait 10s here?

cj-chung:
@yanmo96 Please also add MIT license into these files.

kubernetes/createYaml.sh
yanmo96: It checks the status every 10s

cj-chung (approved):
LGTM

PR #739, opened 2022-02-03
Added in namespace field for ansible deploy script

yanmo96:
This PR added in the "namespace" field for alcor services on k8s.
Originally we don't need the  "namespace" field when we deploy with shell script, since they are all in the default namespace.
Now we want to switch use ansible deployment script, it need the "namespace" field in yaml. Since we can actually switch our "default" to other namespaces.


xieus:
> Now we want to switch use ansible deployment script, it need the "namespace" field in yaml. Since we can actually switch our "default" to other namespaces.

Good and detailed context.

xieus (approved):
LGTM

PR #742, opened 2022-02-05
[DPM] Add gRPC/MQ auto switch mechanism

VanderChen:
In this PR,

1. gRPC/MQ path switch logic is added in PathManagerService
2. a new cache is added, which stores the current chosen path for each VPC.

The gRPC only mode has passed the Jenkins test.

### DIFF
--- a/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
+++ b/services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
@@ -18,18 +18,106 @@
 
 package com.futurewei.alcor.dataplane.service.impl;
 
+import com.futurewei.alcor.dataplane.cache.SubnetPortsCacheV2;
+import com.futurewei.alcor.dataplane.cache.VpcPathCache;
+import com.futurewei.alcor.dataplane.cache.VpcSubnetsCache;
+import com.futurewei.alcor.dataplane.exception.InvalidPathModeException;
 import com.futurewei.alcor.web.entity.port.PortEntity;
+import com.futurewei.alcor.web.entity.subnet.InternalSubnetPorts;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 
 @Service
 public class PathManagerService {
+    @Autowired
+    private VpcSubnetsCache vpcSubnetsCache;
+
+    @Autowired
+    private SubnetPortsCacheV2 subnetPortsCache;
+
+    @Autowired
+    private VpcPathCache vpcPathCache;
+
     private static boolean USE_GRPC = true;
+
+    @Value("${path.mode}")
+    private static String PATH_MODE;
+
+    @Value("${UPPER_VPC_SIZE}")
+    private static int UPPER_VPC_SIZE;
+    @Value("${LOWER_VPC_SIZE}")
+    private static int LOWER_VPC_SIZE;
 
     public boolean isFastPath() {
         return USE_GRPC;
     }
 
-    public boolean isFastPath(PortEntity portEntity) {
-        return USE_GRPC;
+    public boolean isFastPath(PortEntity portEntity) throws Exception{
+        switch (PATH_MODE) {
+            case "GRPC":
+                return USE_GRPC;
+            case "MQ":
+                return !USE_GRPC;
+            case "MIXED":
+                return choosePath(portEntity);
+            default:
+                throw new InvalidPathModeException();
+        }
+    }
+
+    private boolean choosePath(PortEntity portEntity) throws Exception{
+        if (portEntity == null) {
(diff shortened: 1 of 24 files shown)


services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
```diff
@@ -18,18 +18,122 @@
...
-    public boolean isFastPath(PortEntity portEntity) {
-        return USE_GRPC;
+    public boolean isFastPath(PortEntity portEntity) throws Exception{
+        switch (PATH_MODE) {
+            case "GRPC":
+                return USE_GRPC;
+            case "MQ":
+                return !USE_GRPC;
+            case "MIXED":
+                return choosePath(portEntity);
```
cj-chung, line 71: @VanderChen The "Mixed" mode for me seems not a real MIXED mode, doesn't it? I think "AUTO" is better to describe this mode.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
```diff
@@ -18,18 +18,122 @@
...
+        } catch (Exception e) {
+            throw e;
+        }
+
+        // Path switch logic
+        boolean chosenPath;
+
+        if (currentPath == USE_GRPC) {
+            if (numberOfPorts > UPPER_VPC_SIZE) {
+                chosenPath = !USE_GRPC;
```
cj-chung, line 121: @VanderChen Why the line#105 shows that using `USE_GRPC` when `numberOfPorts > UPPER_VPC_SIZE`, but the line#120 shows that using `!USE_GRPC` in the same condition?

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
```diff
@@ -18,18 +18,122 @@
...
+
+        // Get VPC size - number of ports
+        AtomicInteger atomicNumberOfPorts = new AtomicInteger();
+        List<String> subnetIds;
+        subnetIds = vpcSubnetsCache.getVpcSubnets(vpcId).getSubnetIds();
+        subnetIds.stream().forEach(subnetId -> {
+            try {
+                InternalSubnetPorts internalSubnetPorts = subnetPortsCache.getSubnetPorts(subnetId);
+                atomicNumberOfPorts.addAndGet(internalSubnetPorts.getPorts().size());
+            } catch (Exception e) {
```
cj-chung, line 92: @VanderChen  Do we have a better way to retrieve the VPC size (total number of ports) instead of calculating the size on the fly. I am worry about the latency here if it's a very big VPC.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
VanderChen: @cj-chung Line#105 initialize a topic for a new VPC, and line#120 handles the situation of path change. As designed in [doc](https://github.com/futurewei-cloud/alcor/blob/master/docs/modules/ROOT/pages/mq_services/message_queue_system.adoc#channel-selection-algorithm), a simplified algorithm is implemented here.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
VanderChen: @cj-chung Another option is counting in VpcPathCache.  When creating ports, the counter in the cache is incremented.

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
```diff
@@ -18,18 +18,124 @@
...
+        boolean currentPath;
+        try {
+            currentPath = vpcPathCache.getCurrentPathByVpcId(vpcId);
+        } catch (NullPointerException e) {
+            if (numberOfPorts > UPPER_VPC_SIZE) {
+                vpcPathCache.setPath(vpcId, USE_GRPC);
+                return !USE_GRPC;
+            } else {
+                vpcPathCache.setPath(vpcId, !USE_GRPC);
+                return USE_GRPC;
```
cj-chung, line 111: @VanderChen line#107 (`USE_GRPC`) and line #108 is not match (`return !USE_GRPC`). Same as line#110 and line#111.

cj-chung (changes requested):

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/service/impl/PathManagerService.java
cj-chung: @VanderChen Please create an issue for this problem and link to this PR

VanderChen: Issue #744 is created and it seems that I have no authority to link the issue.

cj-chung (approved):
LGTM

PR #751, opened 2022-08-11
Add data path ip used for Arion gateway

DavidLiu506:


services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/grpc/DataPlaneClientImplV2.java
```diff
@@ -306,8 +308,11 @@ public void onCompleted() {
                 finishLatch.countDown();
             }
         };
-
         StreamObserver<Goalstate.GoalStateV2> requestObserver = asyncStub.pushGoalStatesStream(responseObserver);
+        if (arionGatwayEnabled) {
+            goalStateV2 = goalStateV2.toBuilder().clearNeighborStates().build();
+        }
```
xieus, line 314: @DavidLiu506 What does the clearNeighborStates do here?

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
```diff
@@ -60,6 +60,9 @@ public class NodeInfo implements Serializable {
     @QuerySqlField(index = true)
     private String ncm_id;
 
+    @JsonProperty("br-vxlan")
+    private String brVxlan;
```
xieus, line 64: @DavidLiu506 I recommend calling the new field as "dataPathIp" in this PR, and then rename the "localIP" as "controlPathIp" in next PR. 

This way, we could make the IP usage very clear.  

P.S. we may not use vxlan for our data path, so don't title a name to a specific technique.

xieus (commented):
@DavidLiu506 Two comments. Please take a look/

services/data_plane_manager/src/main/java/com/futurewei/alcor/dataplane/client/grpc/DataPlaneClientImplV2.java
DavidLiu506: If ArionGateway enable, neighbor state do not send to ACA to create a neighbor rule because package will route to gate way.

web/src/main/java/com/futurewei/alcor/web/entity/node/NodeInfo.java
```diff
@@ -60,6 +60,9 @@ public class NodeInfo implements Serializable {
     @QuerySqlField(index = true)
     private String ncm_id;
 
+    @JsonProperty("data-path-ip")
+    private String dataPathIp;
+
```
cj-chung, line 65: @DavidLiu506 Do we need `mac_address` for `data-path-ip`? I think the data path should use mac for L2 forwarding, right?

DavidLiu506: ACA don't need it. Arion use mac_address.

cj-chung: @xieus correct me if I am wrong. If ACA doesn't need the mac address, we don't need to figure out how to get mac address from the pod in k8s.

cj-chung: That would be save a lot of time for us :).

DavidLiu506: Yes, you are right.

cj-chung: @DavidLiu506 Just confirm with @xieus Alcor still require `mac_address` for `data path`. We can reuse the current `mac_address` field for `data path`'s mac address.

xieus: As discussed, this is an implementation issue, and Dahai will create an issue to track it so that we will follow up later.

xieus (approved):

PR #766, opened 2022-12-22
Kubernetes Environment: Wait for Ignite Fix

phudtran:
This PR fixes the following:

- Alcor services coming up before ignite resulting in them failing to run.


xieus (approved):
Excellent fix! Thanks @phudtran

